PR #1, opened 2016-04-18
ENH: initial structure, dynamic generation of commands

gregcaporaso:
cc @wasade

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,50 @@
+# q2cli
+A [click-based](http://click.pocoo.org/) command line interface for [QIIME 2](https://github.com/biocore/qiime2).
+
+# Installation and usage
+
+To use this you'll need to have QIIME 2 installed, as well as some plugins to interact with. Since there are not official releases of these packages yet, you can install the development versions as follows. First we'll just get the interface working. *(The `setuptools` specification seems to be required due to an issue described [here](https://github.com/pypa/setuptools/issues/523).)*
+
+```bash
+conda create -n q2cli python=3.5 jupyter scikit-bio 'setuptools<20.5.0' -c biocore
+source activate q2cli
+# install qiime and q2cli
+pip install https://github.com/biocore/qiime2/archive/master.zip https://github.com/qiime2/q2cli/archive/master.zip
+```
+
+You can now run ``qiime --help`` to see the available commands. You can discover what plugins you currently have installed with ``qiime plugins``
+
+```bash
+qiime --help
+qiime --plugins
+qiime --version
+```
+
+There's not much to do without some plugins, so install the ``feature-table`` plugin and execute the ``qiime --plugins`` command again.
+
+```bash
+pip install https://github.com/qiime2/q2-feature-table/archive/master.zip
+qiime --plugins
+```
+
+You should now see that you have one plugin installed.
+
+If you call ``qiime --help`` again, you'll see that you now have a new command available, corresponding to that plugin. To see what subcommands that plugin defines, call it with ``-help``:
+
+```bash
+qiime feature-table --help
+```
+
+Install the ``diversity`` plugin as well. You'll then have two plugins installed, as well as a couple more workflows.
+
+```bash
+pip install https://github.com/qiime2/q2-diversity/archive/master.zip
+qiime --plugins
+qiime diversity --help
+```
+
+Next you should try to run a command that actually does some work. ``qiime diversity feature_table_to_pcoa`` is a good one to try. You can download demo input ``qtf`` files to work with [here](https://github.com/qiime2/q2d3/tree/master/demo/analysis-dir). Here's the command I ran:
+
+```bash
+qiime diversity feature_table_to_pcoa --feature_table table.qtf --phylogeny phylogeny.qtf --metric unweighted_unifrac --depth 50 --distance_matrix uu-dm.qtf --pcoa_results uu-pc.qtf
+```
--- a/q2cli/cli.py
+++ b/q2cli/cli.py
@@ -1,0 +1,146 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2016--, QIIME 2 development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+import click
+import qiime
+from qiime.sdk import PluginManager, SubprocessExecutor
+
+from . import __version__ as q2cli_version
+
+
+CONTEXT_SETTINGS = {'plugin_manager': PluginManager()}
+
+
+class QiimeCLI(click.MultiCommand):
+
+    def list_commands(self, ctx):
+        plugins = ctx.obj['plugin_manager'].plugins.keys()
+        return sorted(plugins)
+
+    def get_command(self, ctx, name):
+        if name in ctx.obj['plugin_manager'].plugins:
+            plugin = ctx.obj['plugin_manager'].plugins[name]
+
+            class PluginCommand(click.MultiCommand):
+
+                def list_commands(self, ctx):
+                    workflows = plugin.workflows.keys()
+                    return sorted(workflows)
+
+                def get_command(self, ctx, name):
+                    workflow = plugin.workflows[name]
+                    return _build_command(name, workflow)
+
+            return PluginCommand(ctx)
+        else:
+            return None
+
+
+# Top-level option handlers
+
+def _echo_version(ctx, name, value):
+    if value:
+        click.echo('QIIME version: %s' % qiime.__version__)
+        click.echo('q2cli version: %s' % q2cli_version)
+
+
+def _echo_plugins(ctx, name, value):
+    if value:
+        installed_plugins = ctx.obj['plugin_manager'].plugins
+        if len(installed_plugins) == 0:
+            click.echo('No plugins are currently installed.\nYou can browse '
+                       'the official QIIME 2 plugins at: '
+                       'https://github.com/qiime2.')
+        else:
+            click.echo('Installed plugins:')
+            for name, plugin in installed_plugins.items():
+                click.echo(' %s %s (%s)' %
+                           (name, plugin.version, plugin.website))
+
+
+def _echo_info(ctx, name, value):
+    if value:
+        _echo_version(ctx, None, True)
+        _echo_plugins(ctx, None, True)
+
+
+@click.command(cls=QiimeCLI, invoke_without_command=True,
+               no_args_is_help=True)
+@click.option('--version', is_flag=True, callback=_echo_version,
+              help='Print the version and exit.', expose_value=False)
+@click.option('--plugins', is_flag=True, callback=_echo_plugins,
+              help='List installed plugins and exit.', expose_value=False)
+@click.option('--info', is_flag=True, callback=_echo_info,
+              help='Print system details and exit.', expose_value=False)
+@click.pass_context
+def cli(ctx):
+    pass
+
+# TODO: update keys to be the types (rather than their str representations)
+# pending biocore/qiime2#46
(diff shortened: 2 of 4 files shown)


q2cli/cli.py
```diff
@@ -0,0 +1,104 @@
...
+import click
+import qiime
+from qiime.sdk import PluginManager, SubprocessExecutor
+
+from . import __version__ as q2cli_version
+
+@click.group()
+@click.pass_context
+def cli(ctx):
+    ctx.obj['plugin_manager'] = PluginManager()
```
wasade: `PluginManager` is instantiated twice, once here and also on line 43. Is that necessary?

q2cli/cli.py
```diff
@@ -0,0 +1,104 @@
...
+
+@cli.command()
+def version():
+    click.echo('QIIME version: %s' % qiime.__version__)
+    click.echo('q2cli version: %s' % q2cli_version)
+
+@cli.command()
+@click.pass_context
+def plugins(ctx):
+    installed_plugins = ctx.obj['plugin_manager'].plugins.items()
```
wasade: Can defer the call to `.items`:

if not len(ctx.obj['plugin_manager'].plugins:
    # die elegantly
    ctx.exit(1)
else:
    # print out plugin deets
    ctx.exit()
```

File: q2cli/cli.py [hunk 3]
```diff
@@ -0,0 +1,104 @@
...
+    if len(installed_plugins) == 0:
+        click.echo('No plugins are currently installed.\nYou can browse the '
+                   'official QIIME 2 plugins at: https://github.com/qiime2.')
+    else:
+        click.echo('Installed plugins:')
+        for name, plugin in installed_plugins:
+            click.echo(' %s %s (%s)' %
+                       (name, plugin.version, plugin.website))
+
+# should the QIIME types be hashable to support this without casting
```
wasade: I'm :+1: on this as these should be singletons anyway

ebolyen:
I love how short this is!

q2cli/cli.py
```diff
@@ -0,0 +1,104 @@
...
+# to a string?
+_type_map = {'Int': click.INT, 'Str': click.STRING, 'Float': click.FLOAT}
+
+def build_commands(group):
+    plugin_manager = PluginManager()
+    for plugin_name, plugin in plugin_manager.plugins.items():
+        for workflow_name, workflow in plugin.workflows.items():
+            parameters = []
+            for ia_name, ia_type in workflow.signature.input_artifacts.items():
```
wasade: Would love a mechanism to cache here. Strongly suspect these blocks here contribute to the large overhead of just running `qiime`. I can see a few options but not worth blocking on.

q2cli/cli.py
```diff
@@ -0,0 +1,104 @@
...
+                                   help=str(ip_type))
+                parameters.append(p)
+            for oa_name, oa_type in workflow.signature.output_artifacts.items():
+                p = click.Option(['--%s' % oa_name],
+                                   required=True,
+                                   type=click.Path(exists=False, dir_okay=False),
+                                   help='Output %s' % str(oa_type))
+                parameters.append(p)
+
+            def create_callback(wf):
```
wasade: is it necessary that this method is define here vs. at module scope? I'm not seeing the ties to local variables but might be missing it.

q2cli/cli.py
```diff
@@ -0,0 +1,104 @@
...
+                def f(**kwargs):
+                    # execute workflow
+                    executor = SubprocessExecutor()
+                    input_artifacts = {
+                        ia_name: kwargs[ia_name]
+                        for ia_name in wf.signature.input_artifacts.keys()}
```
wasade: call to `.keys` is not necessary

q2cli/cli.py
```diff
@@ -0,0 +1,104 @@
...
+                    }
+                    future_ = executor(wf,
+                                       input_artifacts,
+                                       input_parameters,
+                                       output_artifacts)
+
+                    # if in the future we want the cli to not block but rather
+                    # start daemonized processes, we just don't access result
+                    # here, as that's the step that blocks
+                    completed_process = future_.result()
```
wasade: Would it be possible to have a `--daemonize` exposed at the plugin or higher level?

q2cli/cli.py
```diff
@@ -0,0 +1,104 @@
...
+
+                    if completed_process.returncode != 0:
+                        print(completed_process.stdout)
```
wasade: Should that be `stderr`?

wasade: Actually, I think regardless of exit status, that both stdout and stderr should be dumped (via the appropriate file descriptors)?

q2cli/cli.py
```diff
@@ -0,0 +1,104 @@
...
+                        exit(-1)
```
wasade: ctx.exit(1)

wasade: ...basically, click has wrappers around all of these methods. The reason it's advantageous to use them is if you're, say, hooking up the interface commands into a test harness where a call to `sys.exit` raises `SystemExit`, but in a test harness situation, you do not want that to be raised

q2cli/cli.py
```diff
@@ -0,0 +1,104 @@
...
+                return f
+
+            callback = create_callback(workflow)
+            c = click.Command('%s.%s' % (plugin_name, workflow_name),
```
wasade: Re: existing comment on command groups, but we're on the same page there

wasade:
Comments added. Reminds me of pyqi just so much more elegant

q2cli/cli.py
gregcaporaso: Probably not necessary, thanks.

q2cli/cli.py
gregcaporaso: Yeah, good point. I'll create an issue or todo for that.

q2cli/cli.py
gregcaporaso: Nope, probably not, thanks.

q2cli/cli.py
gregcaporaso: I was thinking we'd have a ``--wait/--no-wait`` or something like that. I'll create an issue for this.

q2cli/cli.py
gregcaporaso: Yep, probably right.

q2cli/cli.py
gregcaporaso: I'll add an issue for that too.

q2cli/cli.py
gregcaporaso: I can't seem to get this working right now - I think the group's context isn't active when I call ``build_command``, which might mean that it would make more sense for that to happen as part of ``cli`` itself. I'm just going to leave this as-is for the moment, since it doesn't impact any UI stuff.

q2cli/cli.py
gregcaporaso: Are the exit calls necessary here? I deferred the ``.items`` call, but I think just ``click.echo``ing should be all I need to do.

gregcaporaso:
@wasade, thanks for the input! I got part way through your comments but I'm out of time. I think we're going to spend some time getting this checked in on Friday, so I'll either address the rest of them then, or create issues. 

Let's not merge this PR for the moment - we can merge on Friday.

q2cli/cli.py
```diff
@@ -0,0 +1,113 @@
...
+            completed_process = future_.result()
+
+            if completed_process.returncode != 0:
+                print(completed_process.stdout,
+                      file=sys.stdout)
+                print(completed_process.stderr,
+                      file=sys.stderr)
+                ctx.exit(1)
+        else:
+            # hmmm, I think this is still blocking
```
wasade: it could be that `executor` is blocking still? Would need to trace the logic to be sure...

q2cli/cli.py
wasade: Thanks!

q2cli/cli.py
```diff
@@ -0,0 +1,113 @@
...
+        if ctx.obj['wait']:
+            # block until result is ready
+            completed_process = future_.result()
+
+            if completed_process.returncode != 0:
+                print(completed_process.stdout,
+                      file=sys.stdout)
+                print(completed_process.stderr,
+                      file=sys.stderr)
+                ctx.exit(1)
```
wasade: In retrospect, it might be good to return `completed_process.returncode` which would allow plugins to have an exit status vocabulary

wasade: ...i mean, `ctx.exit(completed_process.returncode)`

q2cli/cli.py
wasade: Okay, no worries at all

wasade:
Sounds good. Thanks again, this is super awesome.

q2cli/cli.py
```diff
@@ -0,0 +1,113 @@
...
+        click.echo('No plugins are currently installed.\nYou can browse the '
+                   'official QIIME 2 plugins at: https://github.com/qiime2.')
+    else:
+        click.echo('Installed plugins:')
+        for name, plugin in installed_plugins.items():
+            click.echo(' %s %s (%s)' %
+                       (name, plugin.version, plugin.website))
+
+# should the QIIME types be hashable to support this without casting
+# to a string?
```
jairideout: 👍

ebolyen: I will make it so.

q2cli/cli.py
jairideout: 👍 for returning the original processes' return code, could be useful for debugging

q2cli/cli.py
jairideout: Just [followed up](https://github.com/qiime2/q2cli/issues/2#issuecomment-212004236) on #2, we should be able to support lazy loading of subcommands. Not blocking for this PR

jairideout:
@gregcaporaso took a quick pass through, this looks great! 🚀

README.md
```diff
@@ -1,2 +1,38 @@
...
+
+# Installation
+
+To use this you'll need to have QIIME 2 installed, as well as some plugins to interact with. Since there are not official releases of these packages yet, you can install the development versions as follows. First we'll just get the interface working. *(The `setuptools` specification seems to be required due to an issue described [here](https://github.com/pypa/setuptools/issues/523).)*
+
+```bash
+conda create -n q2cli python=3.5 jupyter scikit-bio 'setuptools<20.5.0' -c biocore
+source activate q2cli
+# install qiime and q2cli
+pip install https://github.com/biocore/qiime2/archive/master.zip https://github.com/gregcaporaso/q2cli/archive/master.zip
```
gregcaporaso: update q2cli install url

q2cli/cli.py
gregcaporaso: now biocore/qiime2#46

q2cli/cli.py
gregcaporaso: Also, replace ``print`` with ``click.echo``

q2cli/cli.py
gregcaporaso: I don't think that's it - we tested that previously. We'll troubleshoot for another few minutes, and if we're still stuck we'll get it working in another PR.

README.md
gregcaporaso: done

q2cli/cli.py
gregcaporaso: done

q2cli/cli.py
gregcaporaso: #3 is updated with some details on this, we're going to come back to it. It's not trivial unfortunately.

q2cli/cli.py
gregcaporaso: This is addressed in the latest version, which now does lazy loading of the command line interfaces (the interface for a command is only created when that command is called).

wasade:
Congrats on PR #1!

gregcaporaso:
Thanks, and thanks for the input!

PR #11, opened 2016-06-22
adds support for running visualizations

gregcaporaso:


q2cli/cli.py
```diff
@@ -80,22 +91,66 @@ def _echo_info(ctx, name, value):
...
+
+def _build_visualization_callback(wf):
+    # TODO there is a lot of code duplicated between
+    # _build_visualization_callback and _build_workflow_callback - revisit
+    # this after the refactoring that is happening as part of
+    # https://github.com/qiime2/qiime2/issues/39
+    def f(ctx, **kwargs):
+        # TODO nest output_dir under user-configured temporary directory,
+        # pending https://github.com/qiime2/qiime2/issues/12
+        output_dir = tempfile.TemporaryDirectory()
```
ebolyen: Ideally this will be provided by the framework as well, so that the user using the `Artifact` API will just provide the "actual" arguments, and the SDK will fill in the temporary directory implicitly (managed by a configuration)

ebolyen:
Looks good to me :+1:

Issue #17, opened 2016-07-01
add a view command that opens qzv files


gregcaporaso:
This would be really helpful for alpha, and we should be able to do this using the [``webbrowser`` module](https://docs.python.org/3.5/library/webbrowser.html) in Python. I started on this in a local branch, and can finish this one up for alpha.

Issue #16, opened 2016-07-01
include method/visualizer name and description in help text


jairideout:
It'd be great to include a method/visualizer's `name` and `description` in the generated help text for the alpha release.

PR #18, opened 2016-07-08
various enhancements

gregcaporaso:
* Adds ``info`` command which provides ``print_qiime_config.py``-like information
* Adds ``tools`` group which contains ``extract`` and ``view`` (#17) (and will contain ``import`` soon)
* Adds method/visualizer name and description as short_help and help, respectively (#16)
* Removes ``--plugins`` and ``--info`` parameters, as this functionality is now accessible in ``qiime info``.

### DIFF
--- a/q2cli/_info.py
+++ b/q2cli/_info.py
@@ -1,0 +1,69 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2016--, QIIME 2 development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+
+import sys
+
+import pip
+import click
+import qiime
+
+from . import __version__ as q2cli_version
+
+
+def _echo_version(ctx=None, name=None, value=True):
+    if value:
+        pyver = sys.version_info
+        click.echo('Python version: %d.%d.%d' %
+                   (pyver.major, pyver.minor, pyver.micro))
+        click.echo('QIIME version: %s' % qiime.__version__)
+        click.echo('q2cli version: %s' % q2cli_version)
+
+
+def _echo_plugins():
+    plugin_manager = qiime.sdk.PluginManager()
+    installed_plugins = plugin_manager.plugins
+    if len(installed_plugins) == 0:
+        # TODO: update this URL, pending:
+        # https://github.com/qiime2/qiime2/issues/80
+        # https://github.com/qiime2/qiime2/issues/82
+        click.secho('No plugins are currently installed.\nYou can browse '
+                    'the official QIIME 2 plugins at: '
+                    'https://github.com/qiime2')
+    else:
+        for name, plugin in installed_plugins.items():
+            click.echo('%s %s (%s)' %
+                       (name, plugin.version, plugin.website))
+
+
+def _echo_installed_packages():
+    # This code was derived from an example provide here:
+    # http://stackoverflow.com/a/23885252/3424666
+    installed_packages = sorted(["%s==%s" % (i.key, i.version)
+                                for i in pip.get_installed_distributions()])
+    for e in installed_packages:
+        click.echo(e)
+
+
+@click.command(help='Display information about QIIME.')
+@click.option('--py-packages', is_flag=True,
+              help='Display names and versions of all installed Python '
+                   'packages.')
+def info(py_packages):
+    click.secho('System versions', fg='green')
+    _echo_version()
+    click.secho('\nInstalled plugins', fg='green')
+    _echo_plugins()
+    if py_packages:
+        click.secho('\nInstalled Python packages', fg='green')
+        _echo_installed_packages()
+    #TODO: update to import these URLs from the framework, pending
+    # https://github.com/qiime2/qiime2/issues/80
+    click.secho('\nTo get help with QIIME, visit http://help.qiime.org.')
+    click.secho('If you use QIIME in any published work, please cite '
+                'http://www.ncbi.nlm.nih.gov/pubmed/20383131')
--- a/q2cli/_tools.py
+++ b/q2cli/_tools.py
@@ -1,0 +1,82 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2016--, QIIME 2 development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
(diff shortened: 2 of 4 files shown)


q2cli/_tools.py
```diff
@@ -0,0 +1,89 @@
...
+        index_extension = index_extension[1:]
+    try:
+        type_ = qiime.sdk.Result.peek(visualization_path).type
+    except zipfile.BadZipFile:
+        raise click.BadParameter(
+            '%s is not a QIIME Result. Only QIIME Visualizations can be '
+            'viewed.' % visualization_path)
+    # TODO: is there a better way to do this type check (without accessing
+    # qiime.core?)
+    if type_ != qiime.core.type.Visualization:
```
gregcaporaso: is there a better way to do this type check (without accessing ``qiime.core``)?

ebolyen: Not at the moment, we might think about exposing types at something like `qiime.sdk.types` (in fact we would probably move `semantic.py`, `primitive.py` and `visualizations.py`, leaving `grammar.py` in core), but I think we need a larger conversation about package structure post-alpha.

q2cli/_info.py
```diff
@@ -0,0 +1,69 @@
...
+def _echo_plugins():
+    plugin_manager = qiime.sdk.PluginManager()
+    installed_plugins = plugin_manager.plugins
+    if len(installed_plugins) == 0:
+        # TODO: update this URL, pending:
+        # https://github.com/qiime2/qiime2/issues/80
+        # https://github.com/qiime2/qiime2/issues/82
+        click.secho('No plugins are currently installed.\nYou can browse '
+                    'the official QIIME 2 plugins at: '
+                    'https://github.com/qiime2.')
```
ebolyen: no period at the end of the URL

q2cli/_info.py
```diff
@@ -0,0 +1,69 @@
...
+import sys
+
+import pip
+import click
+import qiime
+
+from . import __version__ as q2cli_version
+
+
+def _echo_version(ctx, name, value):
```
ebolyen: make these optional with a default `None`

q2cli/_info.py
```diff
@@ -0,0 +1,69 @@
...
+        click.echo(e)
+
+
+@click.command(help='Display information about QIIME.')
+@click.option('--py-packages', is_flag=True,
+              help='Display names and versions of all installed Python '
+                   'packages.')
+def info(py_packages):
+    click.secho('System versions', fg='green')
+    _echo_version(None, None, True)
```
ebolyen: This is gross, make them optional

q2cli/_tools.py
```diff
@@ -0,0 +1,88 @@
...
+    visualization = qiime.sdk.Visualization.load(visualization_path)
+    index_paths = visualization.get_index_paths()
+    if index_extension not in index_paths:
+        raise click.BadParameter(
+            'No index %s file with is present in the archive. Available index '
+            'extensions are: %s' % (index_extension,
+                                    ', '.join(index_paths.keys())))
+    else:
+        # TODO: we're currently extracting twice here - once when we load, and
+        # again here. How can this be avoided?
```
ebolyen: use `visualization.get_index_paths(relative=False)` instead of extracting.

q2cli/_tools.py
```diff
@@ -0,0 +1,88 @@
...
+                # Entering either ^C or Return will cause output_dir to be
+                # cleaned up.
+                click.prompt('Press the Return key or ^C to quit (your '
+                             'visualization may no longer be accessible)',
+                             default='q', prompt_suffix='. ',
+                             show_default=False)
+
+
+@tools.command(help='Extract a QIIME Arifact or Visualization.')
+@click.argument('path')
```
ebolyen: :+1: for positional arg

q2cli/_tools.py
```diff
@@ -0,0 +1,88 @@
...
+                    "the command exits, use 'qiime extract'.")
+@click.argument('visualization-path')
+@click.option('--index-extension', required=False, default='html',
+              help='The extension of the index file that should be opened. '
+                   '[default: html]')
+def view(visualization_path, index_extension):
+    if index_extension.startswith('.'):
+        index_extension = index_extension[1:]
+    try:
+        type_ = qiime.sdk.Result.peek(visualization_path).type
```
ebolyen: You could use `qiime.sdk.Visualization.peek` instead, it would error if it's an `Artifact`.

q2cli/cli.py
```diff
@@ -6,30 +6,35 @@
...
+    _plugin_manager = PluginManager()
+    _builtin_commands = {'tools': q2cli._tools.tools,
+                         'info': q2cli._info.info}
+
     def list_commands(self, ctx):
-        plugins = PLUGIN_MANAGER.plugins.keys()
-        return sorted(plugins)
+        plugins = list(sorted(self._plugin_manager.plugins.keys()))
+        commands = list(sorted(self._builtin_commands.keys())) + plugins
+        return commands
```
ebolyen: The `+ plugins` is kind of hiding at the end, what if you did:
`return builtins + plugins` which would be more obvious.

ebolyen:
Done reviewing, this is pretty awesome! Couple of places where you could simplify the code, especially using `get_index_paths(relative=False)`

q2cli/_tools.py
gregcaporaso: Awesome, that lets me get around accessing ``qiime.core``. Note that the ``except (zipfile.BadZipFile, KeyError)`` catches passing a non-zip or non-result file.

q2cli/_tools.py
gregcaporaso: 🎸

q2cli/_tools.py
```diff
@@ -0,0 +1,82 @@
...
+        raise click.BadParameter(
+            '%s is not a QIIME Visualization. Only QIIME Visualizations can '
+            'be viewed.' % visualization_path)
+    except TypeError:
+        raise click.BadParameter(
+            '%s is not a Visualization. Only Visualizations can be viewed.' %
+            visualization_path)
+
+    visualization = qiime.sdk.Visualization.load(visualization_path)
+    index_paths = visualization.get_index_paths(relative=False)
```
ebolyen: :+1:

ebolyen:
This is awesome!

PR #21, opened 2016-07-14
Remove types

Kleptobismol:
Stops commands from being displayed if their associated plugin does not have methods or a visualizer

### DIFF
--- a/q2cli/cli.py
+++ b/q2cli/cli.py
@@ -27,6 +27,10 @@ class QiimeCLI(click.MultiCommand):
 
     def list_commands(self, ctx):
         plugins = list(sorted(self._plugin_manager.plugins.keys()))
+        for idx, key in enumerate(plugins):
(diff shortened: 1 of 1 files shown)


q2cli/cli.py
```diff
@@ -27,6 +27,10 @@ class QiimeCLI(click.MultiCommand):
 
     def list_commands(self, ctx):
         plugins = list(sorted(self._plugin_manager.plugins.keys()))
+        for idx, key in enumerate(plugins):
+            plugin = self._plugin_manager.plugins[key]
+            if not plugin.methods or not plugin.visualizers:
```
jairideout: This will remove plugins from the list if they don't have methods or if they don't have visualizers. We want to remove the plugin only if it has exactly zero methods *and* visualizers.

jairideout:
Thanks @Kleptobismol!

PR #26, opened 2016-07-16
ENH: Add i/o prefix to input and output parameters

jakereps:


q2cli/cli.py
```diff
@@ -237,6 +239,15 @@ def _build_visualizer_command(name, visualizer):
                          help=visualizer.description)
 
 
+def _filter_kwargs(kwargs):
```
gregcaporaso: The names used here are a little confusing. What about these changes:

* ``_filter_kwargs`` to ``_option_names_to_api_names`` (filtering kwargs would suggest to me that you're getting rid of some of them) 
* ``stripped_kwarg`` to ``api_name``
* ``stripped_kwargs`` (line 127) to ``option_name_map``

gregcaporaso:
Just one comment on this @jakereps.

jakereps:
Thanks @gregcaporaso, should be good to go!

ebolyen:
:tada:

PR #29, opened 2016-07-16
ENH: Add output directory support for methods and visualizers

jakereps:


q2cli/cli.py
```diff
@@ -207,12 +224,20 @@ def _build_output_option(name, type_):
...
         type=click.Path(exists=False, dir_okay=False),
         help='Output %s' % str(type_[0]))
     return result
 
 
+def _build_output_dir_option():
+    return click.Option(
+        ['--output-dir'],
+        required=False,
+        type=click.Path(exists=False, dir_okay=True),
```
jairideout: Can you add `file_okay=False` to disallow files?

jakereps:
Thanks, @jairideout! Addressed your earlier comment and added the functionality that you, @ebolyen, and I talked about.

q2cli/cli.py
```diff
@@ -217,12 +233,27 @@ def _build_output_option(name, type_):
...
+def _build_output_dir_option():
+    return click.Option(
+        ['--output-dir'],
+        required=False,
+        type=click.Path(exists=False, dir_okay=True, file_okay=False),
+        help='Output all results to a directory')
+
+
+def _build_output_filepath(oa_name, output_dir):
+    oa_value = oa_name
```
gregcaporaso: This can just be:

```python
return os.path.join(output_dir, oa_name)
```

PR #33, opened 2016-07-17
ENH: sort plugin command options alphabetically within inputs, parameters, and output groups

jairideout:
Previous option order was stochastic.

Revert this commit when https://github.com/qiime2/qiime2/issues/70 is fixed.


jairideout:
Huzzah! 🎆

PR #39, opened 2016-07-21
DOC: make metadata/metadata category option help text more general

jairideout:
Now does not have an opinion on the metadata axis it receives.

Pair-programmed with @jakereps.


ebolyen:
:+1: :+1: :+1: :+1: :+1:

PR #42, opened 2016-08-05
ENH: Import urls from qiime.sdk

jakereps:
Utilizes API provided by qiime2/qiime2/issues/80

**Tests depend on either a new QIIME 2 release to conda or swapping QIIME 2 to fetch from the GitHub master branch.**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,9 @@ before_install:
   - ./miniconda.sh -b
   - export PATH=/home/travis/miniconda3/bin:$PATH
 install:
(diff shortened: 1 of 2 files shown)


jairideout:
👍 

The tests are failing because travis is installing the release version of QIIME 2. Can you update .travis.yml to install the latest development version of the framework? Probably makes sense as a separate commit.

jakereps:
Thanks @jairideout, was just about to ping the Slack to ask which one would be a better way to go! Should be good to go!

jairideout:
Thanks @jakereps!

PR #43, opened 2016-08-08
ENH: support discovery of plugin citations and user support text

gregcaporaso:
Here's what this looks like...

``qiime info`` tells you how to access all citations:

screenshot 2016-08-08 13 33 45

which looks like the following: 

screenshot 2016-08-08 13 02 53

(A little bit ugly, but it's a big text dump so I don't really know how to get around it. We can optionally present subsets to the user in the future.)

Calling ``--help`` on a plugin also presents the citation, the website, and how to get help:

screenshot 2016-08-08 13 32 56


gregcaporaso:
This will fail, pending 2.0.2 release of the framework.

PR #46, opened 2016-08-10
Refactor

ebolyen:
Basically a complete rewrite of the generated parts.

@gregcaporaso @jairideout can you review?


q2cli/handlers.py
```diff
@@ -0,0 +1,207 @@
...
+    def get_click_options(self):
+        yield click.Option(
+            ['--' + self.normalized_name],
+            type=click.Path(exists=False, dir_okay=True, file_okay=False),
+            help='Output unspecified results to a directory')
+
+    def get_value(self, arguments, fallback=None):
+        def fallback_(name, normalized_name):
+            path = self._locate_value(arguments, fallback=fallback)
+            os.makedirs(path)
```
ebolyen: This isn't quite working, it is getting called for each output, which `os.makedirs` doesn't like very much...

q2cli/commands.py
```diff
@@ -0,0 +1,173 @@
...
+            handler = self.generated_handlers[name]
+
+            try:
+                value = handler.get_value(kwargs, fallback=fallback)
+            except q2cli.handlers.ValueNotFoundException:
+                click.echo(_OUTPUT_OPTION_ERR_MSG, err=True)
+                click.get_current_context().exit(1)
+
+            if not value.endswith(self.extension):
+                value += self.extension
```
ebolyen: This can be removed now. `ResultHandler.get_value()` does this already.

q2cli/__main__.py
```diff
@@ -0,0 +1,24 @@
...
+# Entry point for CLI
+@click.command(cls=q2cli.commands.RootCommand, invoke_without_command=True,
+               no_args_is_help=True)
+@click.option('--version', is_flag=True, callback=q2cli.info.echo_version,
+              help='Print the version and exit.', expose_value=False)
+def cli():
+    pass
+
+if __name__ is '__main__':
+    cli()
```
ebolyen: The reason for moving the entry-point into this file is we can call:

```bash
python -m q2cli info
```
Which might be useful if a user's environment is totally screwed up.

A `__main__` file is also a pretty obvious place to start looking if you are trying to understand a codebase, so it also felt more natural.

q2cli/commands.py
```diff
@@ -0,0 +1,169 @@
...
+        handler_map = collections.OrderedDict([
+            ('inputs', q2cli.handlers.ArtifactHandler),
+            ('parameters', q2cli.handlers.parameter_handler_factory),
+            ('outputs', functools.partial(q2cli.handlers.ResultHandler,
+                                          self.extension))
+        ])
+
+        for group_type, constructor in handler_map.items():
+            grp = getattr(self.action.signature, group_type)
+            for name, (semtype, _) in sorted(grp.items(), key=lambda x: x[0]):
```
ebolyen: Add back comment indicating to remove sorted when `Signature` preserves API order.

q2cli/handlers.py
```diff
@@ -0,0 +1,214 @@
...
+        super().__init__(name, prefix='m_')
+        self.click_name += '_file'
+
+    def get_click_options(self):
+        yield click.Option(['--' + self.normalized_name],
+                           type=click.Path(exists=True, dir_okay=False),
+                           help='Metadata mapping file')
+
+    def get_value(self, arguments, fallback=None):
+        value = self._locate_value(arguments, fallback)
```
ebolyen: rename to path

q2cli/handlers.py
```diff
@@ -0,0 +1,214 @@
...
+        return qiime.Metadata.load(value)
+
+
+class MetadataCategoryHandler(Handler):
+    def __init__(self, name):
+        self.name = name
+        self.click_names = ['m_%s_file' % name, 'm_%s_category' % name]
+        self.normalized_names = [
+            q2cli.util.normalize_name(n) for n in self.click_names]
+        self.normalize_name = self.normalized_names[0]
```
ebolyen: This won't really work.

q2cli/handlers.py
```diff
@@ -0,0 +1,214 @@
...
+                return fallback(name, normalized_name)
+            except ValueNotFoundException:
+                pass
+
+        # Do we have a default?
+        if self.default is not NoDefault:
+            return self.default
+
+        # Give up
+        raise ValueNotFoundException()
```
ebolyen: Just abort click here.

q2cli/handlers.py
```diff
@@ -0,0 +1,214 @@
...
+
+    def get_type(self):
+        mapping = {
+            'Int': int,
+            'Str': str,
+            'Float': float,
+            'Color': str
+        }
+        predicate = self.ast['predicate']
+        if predicate:
```
ebolyen: Comment that we are only handling this one case and expect this to be changed significantly in the future.

q2cli/__main__.py
ebolyen: check the exit code here

ebolyen:
Notes:

- [ ] Change normalized name -> cli_name
- [ ] docstrings

q2cli/handlers.py
ebolyen: Even better solution completed.

q2cli/handlers.py
ebolyen: Addressed by better solution

ebolyen:
@jairideout ready for re-review

q2cli/__main__.py
ebolyen: it works

jairideout:
Thanks @ebolyen, the rewrite is a huge improvement!!

gregcaporaso:
Awesome, thanks!

PR #48, opened 2016-08-12
ENH: Adds `cmd-config` option to cli

Kleptobismol:
Enables users to specify a config file containing values which will
be used as the options for their command.

### DIFF
--- a/q2cli/handlers.py
+++ b/q2cli/handlers.py
@@ -7,6 +7,8 @@
 # ----------------------------------------------------------------------------
 import os
 import collections
+import configparser
+import warnings
 
 import click
 import qiime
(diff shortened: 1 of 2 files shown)


Kleptobismol:
@jairideout or @ebolyen, ready for review!

q2cli/handlers.py
```diff
@@ -223,7 +258,7 @@ def get_type(self):
         if predicate:
             if predicate['name'] != 'Choices' and self.ast['name'] != 'Str':
                 raise NotImplementedError()
-            return click.Choices(predicate['choices'])
+            return click.Choice(predicate['choices'])
```
ebolyen: This was a typo we discovered. We really just need more unit-tests.

q2cli/handlers.py
```diff
@@ -104,6 +105,43 @@ def fail(*_):
...
+            type=click.Path(exists=True, dir_okay=False, file_okay=True,
+                            readable=True),
+            help='Use config file for command options')
+
+    def get_value(self, arguments, fallback=None):
+        try:
+            path = self._locate_value(arguments, fallback=fallback)
+            config = configparser.ConfigParser()
+            config.read(path)
+            config_group = config['.'.join([self.cli_plugin, self.cli_action])]
```
jairideout: This will fail with (likely) a `KeyError` if the config file doesn't have a matching section for the plugin command. In that case I think it makes sense to raise a `UserWarning` stating that the config file provided via `--cmd-config` does not have a matching section for the specified plugin command.

ebolyen:
Outside of @jairideout's comment this all looks good to me.

q2cli/handlers.py
ebolyen: Actually I guess `config_group` should be renamed to `config_section`.

ebolyen:
Thanks @Kleptobismol!

jairideout:
Thanks @Kleptobismol and @ebolyen!!

Issue #52, opened 2016-08-19
support new qiime.plugin.Bool type


jairideout:
This was recently added to the framework: https://github.com/qiime2/qiime2/pull/116

PR #55, opened 2016-09-07
ENH: added bool as for #52

BenKaehler:
I've added support for Bool primitive in the standard click way. There are other ways to do it, so please let me know if you would like something different.

### DIFF
--- a/q2cli/handlers.py
+++ b/q2cli/handlers.py
@@ -255,7 +255,8 @@ class RegularParameterHandler(GeneratedHandler):
             'Int': int,
             'Str': str,
             'Float': float,
-            'Color': str
+            'Color': str,
+            'Bool': bool
         }
         # TODO: This is a hack because we only support Str % Choices(...) at
         # this point. This entire class should be revisited at some point.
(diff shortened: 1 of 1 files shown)


jairideout:
👍 , this looks good to me but I am not very savvy with the framework's primitive type system. @ebolyen, can you do a final review/merge?

q2cli/handlers.py
```diff
@@ -267,9 +268,16 @@ def get_type(self):
...
+            no_name = self.prefix + 'no_' + self.name
+            cli_no_name = q2cli.util.to_cli_name(no_name)
+            yield click.Option(['--' + self.cli_name + '/--' + cli_no_name])
+        else:
+            yield click.Option(['--' + self.cli_name], type=type)
 
     def get_value(self, arguments, fallback=None):
         value = self._locate_value(arguments, fallback)
+        if self.get_type() is bool:
+            value = self.semtype.encode(value)
```
ebolyen: I would probably just do an early `return value` here instead of encoding to decode.

q2cli/handlers.py
```diff
@@ -267,9 +268,16 @@ def get_type(self):
         return mapping[self.ast['name']]
 
     def get_click_options(self):
-        yield click.Option(['--' + self.cli_name],
-                           type=self.get_type())  # Use the ugly lookup above
+        type = self.get_type()  # Use the ugly lookup above
+        if type is bool:
+            no_name = self.prefix + 'no_' + self.name
+            cli_no_name = q2cli.util.to_cli_name(no_name)
+            yield click.Option(['--' + self.cli_name + '/--' + cli_no_name])
```
ebolyen: Nice!

(granted this handler is a hack, but good extension of it!)

ebolyen:
Thanks @BenKaehler! One tiny comment, but should be good to merge afterwards.

BenKaehler:
Thanks @ebolyen. Done!

jairideout:
👍 , thanks @BenKaehler!

PR #56, opened 2016-09-07
TST: Update test runner for transformers

thermokarst:


thermokarst:
@ebolyen, can you take a look at this when you get a chance? This is blocking #55's tests from passing right now. Thanks!

ebolyen:
Thanks @thermokarst! We really need to figure out some kind of testing harness (it looks like supplying generated artifacts is a common theme).

PR #66, opened 2016-09-30
ENH: more descriptive option help text (e.g. required vs optional)

jairideout:
Displays "[required]" for required inputs and parameters, "[optional]" for parameters with a default value of None, and "[default: <value>]" for parameters with a default value that isn't None.

Additionally, the input and output options are formatted as "Artifact: <type repr>". Output options now describe their usage with --output-dir.

### DIFF
--- a/q2cli/handlers.py
+++ b/q2cli/handlers.py
@@ -15,6 +15,9 @@ import qiime
 
 import q2cli.util
 
+# TODO: revisit how default values are handled when optional artifacts are
+# supported.
+
 # Sentinel to avoid the situation where `None` *is* the default value.
 NoDefault = collections.namedtuple('NoDefault', [])()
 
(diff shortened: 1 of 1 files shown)


q2cli/handlers.py
```diff
@@ -188,43 +192,65 @@ def get_value(self, arguments, fallback=None):
...
-        yield click.Option(['--' + self.cli_names[1]],
-                           type=str,
-                           help='Category from metadata mapping file')
+        md_name = '--' + self.cli_names[0]
+        md_type = click.Path(exists=True, dir_okay=False)
+        md_help = 'Metadata mapping file'
+
+        mdc_name = '--' + self.cli_names[1]
+        mdc_type = str
+        mdc_help = 'Category from metadata mapping file'
```
ebolyen: This is fine, but an alternative would be to `**kwarg` in these arguments, then you would only have to deal with the presence/absence of `default` key and appending `[required]/[optional]` to `help`.

jairideout: Good idea! Addressed in separate commit, be sure to squash that one

PR #67, opened 2016-09-30
ENH: cached CLI state for performance improvements

### LINKED ISSUE #63: fix performance issues to make CLI usable
Related to #6.

jairideout:
Previously, any CLI command would take ~3 seconds (tested with q2-types, q2-feature-table, q2-diversity, q2-emperor installed). The CLI was virtually unusable and tab-completion (as provided out of the box by Click) was even more unusable. This poor performance would have scaled with the number of installed plugins.

Now, the CLI caches its state to a system-specific config directory. The cache is automatically refreshed when a change is made to the current QIIME deployment (e.g. a plugin is uninstalled, a new plugin is installed, a plugin version changes, or the version of q2cli changes).

There are two mechanisms (mainly for developers) to force a cache refresh: the `qiime dev refresh-cache` command, or setting the Q2CLIDEV environment variable to any value. Devs will need to use one of these options to see changes in their plugins (or q2cli itself) reflected in the CLI. Users should not have to force a refresh.

Most commands that don't do any real work take ~420ms to complete, resulting in a ~7x speedup. The performance will have negligible degradation as the number of installed plugins increases. Commands that do real work are faster than before but not benchmarked as their runtimes depend on the specific method being run, and the speedup is less likely to be noticed by users. What's important here is having a snappy CLI while users are exploring its functionality, e.g. requesting help text, running "info" types of commands, etc.

Further optimization is of course possible, and it would be ideal to get the runtime below ~420ms. The speedups here seem to justify closing #63 as the CLI is in a much more usable state now.

Fixes #63.

### DIFF
--- a/q2cli/cache.py
+++ b/q2cli/cache.py
@@ -182,7 +182,7 @@ class DeploymentCache:
         click.secho(
             "QIIME is caching your current deployment for improved "
             "performance. This may take a few moments and should only happen "
-            "once per deployment.", fg='yellow')
+            "once per deployment.", fg='yellow', err=True)
 
         cache_dir = self._cache_dir
 
--- a/q2cli/cache.py
+++ b/q2cli/cache.py
@@ -1,0 +1,281 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2016--, QIIME 2 development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
+
+
+class DeploymentCache:
+    """Cached CLI state for a QIIME deployment.
+
+    In this context, a QIIME deployment is the set of installed Python
+    packages, including their exact versions, that register one or more QIIME 2
+    plugins. The exact version of q2cli is also included in the deployment.
+
+    The deployment cache stores the current deployment's package names and
+    versions in a requirements.txt file under the cache directory. This file is
+    used to determine if the cache is outdated. If the cache is determined to
+    be outdated, it will be refreshed based on the current deployment state.
+    Thus, adding, removing, upgrading, or downgrading a plugin package or q2cli
+    itself will trigger a cache refresh.
+
+    Two mechanisms are provided to force a cache refresh. Setting the
+    environment variable Q2CLIDEV to any value will cause the cache to be
+    refreshed upon instantiation. Calling `.refresh()` will also refresh the
+    cache. Forced refreshing of the cache is useful for plugin and/or q2cli
+    developers who want their changes to take effect in the CLI without
+    changing their package versions.
+
+    Cached CLI state is stored in a state.json file under the cache directory.
+    It is not a public file format and it is not versioned. q2cli is included
+    as part of the QIIME deployment so that the cached state can always be read
+    (or recreated as necessary) by the currently installed version of q2cli.
+
+    This class is intended to be a singleton because it is responsible for
+    managing the on-disk cache. Having more than one instance managing the
+    cache has the possibility of two instances clobbering the cache (e.g. in a
+    multithreaded/multiprocessing situation). Also, having a single instance
+    improves performance by only reading and/or refreshing the cache a
+    single time during its lifetime. Having two instances could, for example,
+    trigger two cache refreshes if Q2CLIDEV is set. To support these use-cases,
+    a module-level `CACHE` variable stores a single instance of this class.
+
+    """
+
+    # Public API
+
+    def __init__(self):
+        import os
+
+        # Indicates if the cache has been refreshed. For performance purposes,
+        # the cache is only refreshed a single time (at maximum) during the
+        # object's lifetime. Thus, "hot reloading" isn't supported, but this
+        # shouldn't be necessary for the CLI.
+        self._refreshed = False
+
+        self._cache_dir = self._get_cache_dir()
+
+        refresh = 'Q2CLIDEV' in os.environ
+        self._state = self._get_cached_state(refresh=refresh)
+
+    @property
+    def plugins(self):
+        """Decoded JSON object representing CLI state on a per-plugin basis."""
+        return self._state['plugins']
+
+    def refresh(self):
+        """Trigger a forced refresh of the cache.
+
+        If the cache has already been refreshed (either by this method or at
+        some point during instantiation), this method is a no-op.
+
+        """
+        if not self._refreshed:
+            self._state = self._get_cached_state(refresh=True)
+
+    # Private API
+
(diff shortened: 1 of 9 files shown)


jairideout:
@ebolyen, I made the couple of fixes you identified yesterday, and fixed the Travis build. This is ready for review 💵

q2cli/info.py
```diff
@@ -6,41 +6,41 @@
...
                     'the official QIIME 2 plugins at: '
-                    'http://2.qiime.org/Plugins')
-    else:
-        for name, plugin in installed_plugins.items():
-            click.echo('%s %s' % (name, plugin.version))
+                    '%s/Plugins' % qiime.sdk.HELP_URL)
 
 
 def _echo_installed_packages():
+    import pip
```
ebolyen: Not important for this PR, but we should switch to using `pkg_resources.working_set` for this.

q2cli/cache.py
```diff
@@ -0,0 +1,281 @@
...
+
+    def _cache_current_state(self, requirements):
+        import json
+        import os.path
+        import click
+
+        click.secho(
+            "QIIME is caching your current deployment for improved "
+            "performance. This may take a few moments and should only happen "
+            "once per deployment.", fg='yellow')
```
ebolyen: Should this be `stderr`? I'm honestly not sure, but it strikes me as something you could filter out if you were inspecting an `info` call.

It's a shame there isn't a `stdinfo` fd, that would be perfect for these kinds of things.

ebolyen: Here's an opinion from the internet: http://stackoverflow.com/a/4919110/579416

q2cli/cache.py
```diff
@@ -0,0 +1,281 @@
...
+    single time during its lifetime. Having two instances could, for example,
+    trigger two cache refreshes if Q2CLIDEV is set. To support these use-cases,
+    a module-level `CACHE` variable stores a single instance of this class.
+
+    """
+
+    # Public API
+
+    def __init__(self):
+        import os
```
ebolyen: Given that `CACHE` executes in this file, we may as well move this import to the top. This will prevent multiple scans to find `os`.

q2cli/cache.py
```diff
@@ -0,0 +1,281 @@
...
+
+        return state
+
+    def _get_plugin_state(self, plugin):
+        state = {
+            # TODO this conversion also happens in the framework
+            # (qiime/plugins.py) to generate an importable module name from a
+            # plugin's `.name` attribute. Centralize this knowledge in the
+            # framework, ideally as a machine-friendly plugin ID (similar to
+            # `Action.id`).
```
ebolyen: :+1:

q2cli/cache.py
```diff
@@ -0,0 +1,281 @@
...
+
+        # Now that the cache is up-to-date, read it.
+        try:
+            with open(state_path, 'r') as fh:
+                return json.load(fh)
+        except json.JSONDecodeError:
+            # 4) The cached state file can't be read as JSON.
+            self._cache_current_state(current_requirements)
+            with open(state_path, 'r') as fh:
+                return json.load(fh)
```
ebolyen: If this fails, it might be worth letting the user know just how *sorry we are*, and that there's pretty much nothing they can to fix it.

q2cli/cache.py
```diff
@@ -0,0 +1,286 @@
...
+        # 1) We have been explicitly told to refresh.
+        # 2) The current deployment requirements are different than the cached
+        #    requirements.
+        if refresh or (current_requirements !=
+                       self._get_cached_requirements()):
+            self._cache_current_state(current_requirements)
+
+        path = os.path.join(self._cache_dir, 'state.json')
+
+        # 3) The cached state file does not exist.
```
ebolyen: Sorry, this one was from an older review, disregard.

q2cli/cache.py
jairideout: All imports are cached after they are first imported, whether they're at module scope or in a function/method, so performance shouldn't be different one way or the other. I just moved all imports (that could be moved) into the methods that needed them, mainly for consistency and in case the code gets used in different ways in the future (e.g. a method becomes deferred where it is currently not deferred). If there was a tie (e.g. with this import), I chose to put them in method scope.

q2cli/cache.py
jairideout: I'm not sure how it would fail at this point, since the cache has been recreated on line 115. It could fail if we wrote invalid JSON while regenerating the cache, but I think we can assume the cache code works as expected?

q2cli/cache.py
jairideout: I guess there is some overhead if the function is called repeatedly, even though the imports are cached. It's probably a wash for this one (because this gets instantiated exactly one time), but I'll move it out (I don't have much of an opinion on this one).

q2cli/handlers.py
```diff
@@ -86,12 +81,17 @@ def __init__(self):
         super().__init__('output_dir')
 
     def get_click_options(self):
+        import click
```
jairideout: Good catch, I'll move this one out.

jairideout:
Thanks for the review @ebolyen! I just noticed my comment asking you to review was posted 3x... not sure how that happened. Sorry you got pinged that many times asking for a review :)

q2cli/cache.py
jairideout: I moved the imports you noted (here and below) and didn't see a change in performance. Okay if I just leave them where they are for consistency?

ebolyen: @jairideout I'm good with consistency!

q2cli/cache.py
ebolyen: That's mostly the case I'm thinking about, but it is fine to leave as is.

q2cli/handlers.py
ebolyen: ¯\\_(ツ)_/¯

ebolyen:
@jairideout I actually didn't get pinged 3 times, might be a UI bug on your side?

ebolyen:
Thanks @jairideout! :tada:

q2cli/cache.py
jairideout: Sounds good. There are a few other things that could make the cache fail with an uninformative error message, but probably not worth fixing unless they happen in the wild (they're all pretty unlikely cases).

q2cli/cache.py
jairideout: Sounds good, I'll leave the imports as-is

jairideout:
Thanks for reviewing and merging @ebolyen! Yeah it looks like a UI bug on my end, it ended up displaying the same comment 6 times when I last checked. A page refresh fixed it :)

PR #77, opened 2016-10-03
BUG: fixes broken boolean flag handler

gregcaporaso:
Processing of boolean flags was broken in master. I discovered this by running ``qiime dada2 denoise`` and getting the following:

```
raise TypeError('Got secondary option for non boolean flag.')
```

The [docs](http://click.pocoo.org/5/options/#boolean-flags) don't set `type` when creating boolean flags with secondary options (i.e., ``--something/--no-something``), so I followed that example.

### DIFF
--- a/q2cli/handlers.py
+++ b/q2cli/handlers.py
@@ -332,17 +332,22 @@ class RegularParameterHandler(GeneratedHandler):
             no_name = self.prefix + 'no_' + self.name
             cli_no_name = q2cli.util.to_cli_name(no_name)
             name = '--' + self.cli_name + '/--' + cli_no_name
+            # click.Option type is determined implicitly for flags with
+            # secondary options, and explicitly passing type=bool results in a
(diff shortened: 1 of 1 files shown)


q2cli/handlers.py
```diff
@@ -336,12 +336,12 @@ def get_click_options(self):
             name = '--' + self.cli_name
 
         if self.default is NoDefault:
-            yield click.Option([name], type=type, help='[required]')
```
jairideout: We still want to specify the type for non-boolean types. I think you'll need to special-case boolean here

gregcaporaso: Sorry! I misread the code, and thought we were in the `if type is bool` block here.

gregcaporaso: That's fixed.

jairideout: No problem, thanks for fixing!

jairideout:
Thanks @gregcaporaso!

PR #78, opened 2016-10-03
ENH/BUG: add --verbose flag; fix two --cmd-config bugs

### LINKED ISSUE #64: CLI should suppress stdout/stderr by default
with option(s) to override this behavior.

jairideout:
Adds a `--verbose` flag to plugin action commands (off by default) that displays any stdout/stderr generated by an action.

Also fixed two bugs in --cmd-config option:

- supplying a value in config was ignored if the option had a default value
- boolean flags in config did not work correctly; any value was evaluated as true except for the empty value, and the value (as a `str`) was forwarded to the action instead of being converted to a `bool` type.

Also improved error message for when `--output-dir` directory already exists (previously was a traceback, now is a nicely formatted Click echo).

Fixes #64.

### DIFF
--- a/q2cli/commands.py
+++ b/q2cli/commands.py
@@ -106,6 +106,7 @@ class ActionCommand(click.Command):
         self.plugin = plugin
         self.action = action
         self.generated_handlers = self.build_generated_handlers()
+        self.verbose_handler = q2cli.handlers.VerboseHandler()
         # Meta-Handlers:
         self.output_dir_handler = q2cli.handlers.OutputDirHandler()
         self.cmd_config_handler = q2cli.handlers.CommandConfigHandler(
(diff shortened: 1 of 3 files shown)


q2cli/commands.py
```diff
@@ -180,7 +185,16 @@ def __call__(self, **kwargs):
         actions_module = importlib.import_module(module_path)
         action = getattr(actions_module, self.action['id'])
 
-        results = action(**arguments)
+        stdout = os.devnull
+        stderr = os.devnull
+        if verbose:
+            # `qiime.util.redirected_stdio` defaults to stdout/stderr when
```
jairideout: Another way to do this is to only invoke the context manager if not in verbose mode. This would avoid redirecting stdio when in verbose mode. On the fence, it probably doesn't matter much either way.

ebolyen: I like the current implementation I think.

jairideout: Cool, let's keep as-is then

PR #84, opened 2016-10-07
ENH: easier tab completion activation: `source tab-qiime`

jairideout:
Removes the ugly and hard-to-remember `eval "$(register-qiime-completion 2> /dev/null)"` command in favor of `source tab-qiime`. Also enables tab completion out of the box for a fresh q2cli install once users run this command.

Thanks @thermokarst and @gregcaporaso for your input!


jairideout:
Thanks @thermokarst!!! 🅰️ ➕

PR #90, opened 2016-10-24
ENH: Add qiime tools export

jakereps:
Adds `qiime tools export --input-path PATH --output-path PATH [--format QIIME2Format]`

Relies on qiime2/qiime2/pull/172

### DIFF
--- a/q2cli/tools.py
+++ b/q2cli/tools.py
@@ -35,7 +35,7 @@ def export_data(input_path, output_path):
 
 @tools.command(name='import', short_help='Import data.',
                help="Import data to create a new QIIME Artifact. See "
-                    "http://2.qiime.org/Importing-data for usage examples "
+                    "https://docs.qiime2.org/2.0.5/import/ for usage examples "
                     "and details on the file types and associated semantic "
                     "types that can be imported.")
 @click.option('--type', required=True,
--- a/q2cli/tools.py
+++ b/q2cli/tools.py
@@ -16,6 +16,23 @@ def tools():
     pass
 
 
+@tools.command(name='export', short_help='Export data.',
+               help="Export data from a QIIME Artifact or Visualization.")
+@click.option('--input-path', required=True,
+              type=click.Path(exists=True, dir_okay=True),
+              help='Path to Artifact or Visualization that should be '
+                   'exported.')
+@click.option('--output-path', required=True,
+              type=click.Path(exists=False, dir_okay=False),
+              help='Path where Result data should be written.')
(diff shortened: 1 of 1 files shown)


jakereps:
Failing tests related to qiime2/qiime2/pull/171

q2cli/tools.py
```diff
@@ -16,6 +16,26 @@ def tools():
     pass
 
 
+@tools.command(name='export', short_help='Export data.',
+               help="Export data from a QIIME Artifact or Visualization. See "
```
thermokarst: Maybe we can just truncate this help text to "Export data from a QIIME Artifact or Visualization.", since any artifact can be exported, and there is no real config or options to provide.

q2cli/tools.py
```diff
@@ -16,6 +16,26 @@ def tools():
...
+    import qiime.sdk
+
+    result = qiime.sdk.Result.load(input_path)
+
+    return result.export_data(output_path)
+
+
 @tools.command(name='import', short_help='Import data.',
                help="Import data to create a new QIIME Artifact. See "
                     "http://2.qiime.org/Importing-data for usage examples "
```
thermokarst: Since `http://2.qiime.org/Importing-data` doesn't exist anymore, can we update the URL to `https://docs.qiime2.org/2.0.5/import/`, while you are making changes above?

jakereps:
Should be good to go, thanks @thermokarst!

q2cli/tools.py
```diff
@@ -16,9 +16,26 @@ def tools():
     pass
 
 
+@tools.command(name='export', short_help='Export data.',
+               help="Export data from a QIIME Artifact or Visualization.")
+@click.option('--input-path', required=True,
+              type=click.Path(exists=True, dir_okay=True),
```
jairideout: I think this should be `dir_okay=False` since inputs will always be .qza/.qzv

q2cli/tools.py
```diff
@@ -16,9 +16,26 @@ def tools():
...
+              help='Path to Artifact or Visualization that should be '
+                   'exported.')
+@click.option('--output-path', required=True,
```
jairideout: Just had some discussion about the `qiime tools export` interface and we decided to have it match `tools extract` pretty closely. I'll issue a PR soon with those changes and will address my comment above about `dir_okay=False`.

jairideout: Changes mentioned here and above are addressed in https://github.com/qiime2/q2cli/pull/94

PR #92, opened 2016-10-25
TST: Fix missing provenance from test setup

jakereps:
Provenance added a parameter to `_from_view()`. Imported and added  required `ImportProvenanceCapture` to get tests to pass.


jairideout:
Thanks @jakereps!

PR #94, opened 2016-10-31
ENH: match interfaces for `tools export` and `tools extract` commands

jairideout:
`tools export` now matches the interface of `tools extract` since they perform similar operations. Updated, expanded, and normalized help text for `tools import`, `tools export`, and `tools extract` as they are all related operations.

### DIFF
--- a/q2cli/tools.py
+++ b/q2cli/tools.py
@@ -37,7 +37,8 @@ def export_data(path, output_dir):
     result.export_data(output_dir)
 
 
-@tools.command(name='import', short_help='Import data into a QIIME Artifact.',
+@tools.command(name='import',
+               short_help='Import data into a new QIIME Artifact.',
                help="Import data to create a new QIIME Artifact. See "
                     "https://docs.qiime2.org/ for usage examples and details "
                     "on the file types and associated semantic types that can "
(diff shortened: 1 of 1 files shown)


q2cli/tools.py
```diff
@@ -16,28 +16,32 @@ def tools():
...
 
-    result = qiime.sdk.Result.load(input_path)
-
-    return result.export_data(output_path)
+    result = qiime.sdk.Result.load(path)
+    result.export_data(output_dir)
 
 
-@tools.command(name='import', short_help='Import data.',
+@tools.command(name='import', short_help='Import data into a QIIME Artifact.',
```
thermokarst: Since we are mirroring `export` above, maybe we should clarify that this is a *new* QIIME Artifact? If someone is unfamiliar with QIIME 2 they might assume that you can import items into an existing artifact.

Options:
- `Import data into a new QIIME Artifact`
- `Create a new QIIME Artifact by importing data.`
- ???

q2cli/tools.py
```diff
@@ -134,11 +138,16 @@ def view(visualization_path, index_extension):
                     break
 
 
-@tools.command(help='Extract a QIIME Artifact or Visualization.')
+@tools.command(short_help="Extract a QIIME Artifact or Visualization archive.",
+               help="Extract all contents of a QIIME Artifact or "
+                    "Visualization's archive. Use 'qiime tools export' to "
```
thermokarst: How about we clarify that there are extra "things" involved in an archive, beside just the source data, that way we can further differentiate from `export`. Maybe something like: `... Visualization's archive, including provenance, metadata, and actual data. Use ...`?

jairideout:
Thanks @thermokarst!

PR #108, opened 2016-12-12
ENH: Add parameter help text support

jakereps:
Adds help text functionality on all methods and visualizers.

Relies on qiime2/qiime2/pull/196

### DIFF
--- a/q2cli/handlers.py
+++ b/q2cli/handlers.py
@@ -104,7 +104,7 @@ class Handler:
             ctx.exit(1)
 
     def _add_description(self, option):
-        if self.description is not NoDefault:
+        if self.description != '':
             option.help += '\n%s' % self.description
         return option
 
--- a/q2cli/handlers.py
+++ b/q2cli/handlers.py
@@ -20,12 +20,14 @@ class ValueNotFoundException(Exception):
 
 
 class Handler:
-    def __init__(self, name, prefix='', default=NoDefault):
+    def __init__(self, name, prefix='', default=NoDefault,
+                 description=NoDefault):
         # e.g. my_option_name
         self.name = name
         # e.g. p_my_option_name
         self.click_name = prefix + name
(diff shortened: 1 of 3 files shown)


q2cli/handlers.py
```diff
@@ -101,6 +103,11 @@ def _parse_boolean(self, string):
             ctx = click.get_current_context()
             ctx.exit(1)
 
+    def _add_description(self, option):
+        if self.description != '':
```
jairideout: Based on this comparison, I think `description=''` makes more sense than `description=NoDefault`. The `NoDefault` sentinel is only necessary for default values where `None` is a valid default.

I suppose `description=None` would also work, since `None` is serialized to JSON `null`. I don't have strong feelings about `None` vs `''` though.

jakereps: 👍 Swapped to `None`. Makes more "Pythonic" sense instead of empty strings.

q2cli/handlers.py
```diff
@@ -301,7 +319,7 @@ def get_value(self, arguments, fallback=None):
 
 
 class MetadataCategoryHandler(Handler):
-    def __init__(self, name, default=NoDefault):
+    def __init__(self, name, default=NoDefault, description=None):
```
jairideout: Should this pass `description` through `super` like `MetadataHandler`?

jairideout:
@jakereps I had one more question, sorry I didn't notice it in my last review.

q2cli/handlers.py
jakereps: Nice catch! I think it definitely should, as I passed it along for all the other ones. Missed one I guess...

jakereps:
Thanks @jairideout! Took two attempts fixing it on mobile, but it's good to go! Haha.

jairideout:
:taco: 👍

PR #110, opened 2016-12-15
ENH: Add feedback to result saving

### LINKED ISSUE #98: Confirm output dir/artifact/viz
This came up at the PHX workshop --- an option to echo the written paths (`qza`s and `qzv`s) to stdout would be more illuminating.

jakereps:
Resolves #98 

Thoughts on verbiage? `Saved to:` `Written to:` `<filepath only>`. Extend the filepath beyond the relative path that is returned by `.save()`?

```
(qiime2) [17:09] qiime2-moving-pictures-tutorial $ qiime diversity beta-group-significance --i-distance-matrix cm1441/unweighted_unifrac_distance_matrix.qza --m-metadata-file sample-metadata.tsv --m-metadata-category BodySite --o-visualization cm1441/unweighted-unifrac-body-site-significance
Saved to: cm1441/unweighted-unifrac-body-site-significance.qzv
(qiime2) [17:09] qiime2-moving-pictures-tutorial $ 
```

### DIFF
--- a/q2cli/commands.py
+++ b/q2cli/commands.py
@@ -215,7 +215,8 @@ class ActionCommand(click.Command):
             results = action(**arguments)
 
         for result, output in zip(results, outputs):
-            result.save(output)
+            click.echo('Saved to: %s' % result.save(output))
+
 
     def handle_in_params(self, kwargs):
         import itertools
--- a/q2cli/commands.py
+++ b/q2cli/commands.py
@@ -215,8 +215,8 @@ class ActionCommand(click.Command):
             results = action(**arguments)
 
         for result, output in zip(results, outputs):
-            click.echo('Saved to: %s' % result.save(output))
-
+            click.secho('Saved %s to: %s' % (result.type,
+                                             result.save(output)), fg='green')
 
     def handle_in_params(self, kwargs):
         import itertools


jairideout:
> Thoughts on verbiage? 

👍, I like the current wording. Do you think it'd be helpful to distinguish between artifacts and visualizations? i.e. `Saved artifact to: foo/bar.qza` and `Saved visualization to: baz.qzv`

What do you think about echoing using green text? That way it's easy to spot, plus it's a "successful" operation.

Do you think this should always echo (like it is now)? Or should `--verbose` activate it?

> Extend the filepath beyond the relative path that is returned by .save()?

I like the current behavior because it'll be relative if the user provided a relative path, and absolute otherwise.

@thermokarst @ebolyen, interested in your thoughts.

thermokarst:
I like differentiating between viz and artifact in the message, also formatting the text green sounds great. I also like the current behavior, echoing whatever the user provided as far as path. Maybe `--verbose` could expand relative to absolute, and pass through absolute as-is?

jakereps:
Thanks @jairideout and @thermokarst, here's the updated output:
image

jairideout:
👍 , thanks @jakereps !

PR #113, opened 2017-01-01
BUG: avoid outdated cache on plugin loading failure

### LINKED ISSUE #88: incomplete cache if plugin loading fails
This came up during some testing I did for the Iceland workshop. If a QIIME 2 plugin fails to load when the cache is being refreshed (and written to disk), the requirements.txt file will be created but state.json will not. Thus, subsequent q2cli commands will fail to detect that a cache refresh is necessary, and installed plugins will not be visible from the CLI.

When I was testing, a plugin relying on matplotlib failed to load due to a matplotlib `ImportError` (the PyQT conda package bug we've all ran into by now). I believe @nbokulich faced a similar issue recently.

For now, this is easily worked around by fixing the offending plugin's installation error and running `qiime dev refresh-cache` (or just delete your cache directory).

This is an easy fix, assigning to myself.

jairideout:
Cache requirements file is now written after all other parts of the cache to avoid creating a newer version of the requirements that doesn't match the previous state.json.

This fixes the case where a user ran q2cli with a functioning QIIME deployment (thus creating a valid cache), and then installed a new plugin and ran q2cli again. A cache refresh was triggered, and if the plugin failed to load for whatever reason, a new requirements file had already been written, and this newer requirements file was out-of-sync with the state.json file that existed from the previous deployment. Thus, the new plugin wouldn't display in the CLI even if the plugin loading problems were resolved because the cache wouldn't be refreshed.

Fixes #88.

### DIFF
--- a/q2cli/cache.py
+++ b/q2cli/cache.py
@@ -193,15 +193,6 @@ class DeploymentCache:
             "once per deployment.", fg='yellow', err=True)
 
         cache_dir = self._cache_dir
-
-        path = os.path.join(cache_dir, 'requirements.txt')
-        with open(path, 'w') as fh:
-            for req in requirements:
-                # `str(Requirement)` is the recommended way to format a
-                # `Requirement` that can be read with `Requirement.parse`.
-                fh.write(str(req))
-                fh.write('\n')
-
         state = self._get_current_state()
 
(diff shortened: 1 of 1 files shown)


jairideout:
Thanks for testing it locally @thermokarst! :1st_place_medal: :taco:

PR #122, opened 2017-02-08
BUG: Not all macOS machines have a $DISPLAY set

### LINKED ISSUE #121: BUG: `qiime tools view` checks for $DISPLAY which isn't set on OS X
We should check if `uname` is `Darwin` and if so, assume we have a windowed environment.

jakereps:
Fixes #121


jairideout:
👍 ! @thermokarst pointed out that there are headless Mac servers out there but I'm not sure how we'd detect that. I'm fine with merging this since it'll solve the problem in most cases. If it's an issue for headless Mac server users in the future we can revisit.

PR #131, opened 2017-03-29
attempt to add --show-importable-types option

andrewsanchez:
I expected to get the set of importable types via `qiime2.sdk.PluginManager.importable_types`, but that just results in a property object.

```bash
(qiime2-dev) [aas229@wind ~/projects/qiime_dev/q2cli ]$ qiime tools import --show-importable-types
<property object at 0x7f1adf7b8318>
```

How can I actually access the set of importable types?

### DIFF
--- a/q2cli/tests/test_cli.py
+++ b/q2cli/tests/test_cli.py
@@ -70,13 +70,12 @@ class CliTests(unittest.TestCase):
         result = self.runner.invoke(
             tools, ['import', '--show-importable-types'])
         self.assertEqual(result.exit_code, 0)
-        self.assertTrue('IntSequence1' in result.output)
-        self.assertTrue('FourInts' in result.output)
-        self.assertTrue('IntSequence1' in result.output)
-        self.assertTrue('IntSequence2' in result.output)
-        self.assertTrue('Kennel[Cat]' in result.output)
-        self.assertTrue('Kennel[Dog]' in result.output)
-        self.assertTrue('Mapping' in result.output)
+
+        import qiime2.sdk
+        importable_types = sorted(qiime2.sdk.PluginManager().importable_types,
+                                key=repr)
+        for name in importable_types:
+            self.assertTrue(name in result.output)
 
     def test_extract(self):
         result = self.runner.invoke(
--- a/q2cli/tests/test_cli.py
+++ b/q2cli/tests/test_cli.py
@@ -70,7 +70,6 @@ class CliTests(unittest.TestCase):
         result = self.runner.invoke(
             tools, ['import', '--show-importable-types'])
         self.assertEqual(result.exit_code, 0)
-        self.assertTrue('IntSequence1' in result.output)
         self.assertTrue('FourInts' in result.output)
         self.assertTrue('IntSequence1' in result.output)
         self.assertTrue('IntSequence2' in result.output)
--- a/q2cli/tests/test_cli.py
+++ b/q2cli/tests/test_cli.py
@@ -67,14 +67,16 @@ class CliTests(unittest.TestCase):
         self.assertFalse('mapping_viz' in commands)
 
     def test_show_importable_types(self):
-        import qiime2.sdk
         result = self.runner.invoke(
             tools, ['import', '--show-importable-types'])
         self.assertEqual(result.exit_code, 0)
-        importable_types = sorted(qiime2.sdk.PluginManager().importable_types,
-                                  key=repr)
-        for item in importable_types:
-            self.assertTrue(item.name in result.output)
+        self.assertTrue('IntSequence1' in result.output)
+        self.assertTrue('FourInts' in result.output)
+        self.assertTrue('IntSequence1' in result.output)
+        self.assertTrue('IntSequence2' in result.output)
+        self.assertTrue('Kennel[Cat]' in result.output)
+        self.assertTrue('Kennel[Dog]' in result.output)
+        self.assertTrue('Mapping' in result.output)
 
     def test_extract(self):
         result = self.runner.invoke(
--- a/q2cli/tests/test_cli.py
+++ b/q2cli/tests/test_cli.py
@@ -70,6 +70,13 @@ class CliTests(unittest.TestCase):
         result = self.runner.invoke(
             tools, ['import', '--show-importable-types'])
         self.assertEqual(result.exit_code, 0)
+        self.assertTrue('IntSequence1' in result.output)
+        self.assertTrue('FourInts' in result.output)
+        self.assertTrue('IntSequence1' in result.output)
+        self.assertTrue('IntSequence2' in result.output)
+        self.assertTrue('Kennel[Cat]' in result.output)
+        self.assertTrue('Kennel[Dog]' in result.output)
+        self.assertTrue('Mapping' in result.output)
 
     def test_extract(self):
         result = self.runner.invoke(
--- a/q2cli/tests/test_cli.py
+++ b/q2cli/tests/test_cli.py
@@ -67,15 +67,14 @@ class CliTests(unittest.TestCase):
         self.assertFalse('mapping_viz' in commands)
 
     def test_show_importable_types(self):
+        import qiime2.sdk
         result = self.runner.invoke(
             tools, ['import', '--show-importable-types'])
         self.assertEqual(result.exit_code, 0)
-
-        import qiime2.sdk
         importable_types = sorted(qiime2.sdk.PluginManager().importable_types,
-                                key=repr)
-        for name in importable_types:
-            self.assertTrue(name in result.output)
+                                  key=repr)
+        for item in importable_types:
+            self.assertTrue(item.name in result.output)
 
     def test_extract(self):
         result = self.runner.invoke(
--- a/q2cli/tests/test_cli.py
+++ b/q2cli/tests/test_cli.py
@@ -66,6 +66,11 @@ class CliTests(unittest.TestCase):
         self.assertFalse('split_ints' in commands)
         self.assertFalse('mapping_viz' in commands)
 
+    def test_show_importable_types(self):
+        result = self.runner.invoke(
+            tools, ['import', '--show-importable-types'])
+        self.assertEqual(result.exit_code, 0)
+
     def test_extract(self):
         result = self.runner.invoke(
             tools, ['extract', self.artifact1_path, '--output-dir',
--- a/q2cli/tools.py
+++ b/q2cli/tools.py
@@ -36,18 +36,21 @@ def export_data(path, output_dir):
     result = qiime2.sdk.Result.load(path)
     result.export_data(output_dir)
 
+
 def show_importable_types(ctx, param, value):
     import qiime2.sdk
 
     PluginManager = qiime2.sdk.PluginManager()
     set_of_importable_types = PluginManager.importable_types
-    set_of_importable_types = sorted([repr(t)for t in  set_of_importable_types])
+    set_of_importable_types = [repr(t) for t in set_of_importable_types]
+    set_of_importable_types = sorted(set_of_importable_types)
 
     if not value or ctx.resilient_parsing:
         return
     for name in set_of_importable_types:
         click.echo(name)
     ctx.exit()
+
 
 @tools.command(name='import',
                short_help='Import data into a new QIIME 2 Artifact.',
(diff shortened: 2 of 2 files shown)


thermokarst:
> How can I actually access the set of importable types?

Hi @andrewsanchez! It looks like you haven't instantiated the plugin manager. An example of that [can be seen here](https://github.com/qiime2/q2cli/blob/master/q2cli/cache.py#L238). So, once you have an instance of the plugin manager available to you, you should be able to grab that `importable_types` property:

```python
plugin_manager = qiime2.sdk.PluginManager()
print(plugin_manager.importable_types)

{FeatureData[PairedEndSequence], DeblurStats, SampleData[AlphaDiversity], ...}
```

Hope that helps!

andrewsanchez:
Hi @thermokarst!  Thanks for the correction.  After instantiating the PluginManager as you showed, I get an empty set:
```python
In [2]: plugin_manager = qiime2.sdk.PluginManager()
In [3]: print(plugin_manager.importable_types)
set()
```

😕

thermokarst:
Hi @andrewsanchez, do you have any [plugins](https://docs.qiime2.org/2017.2/plugins/available/) installed? If not, then the empty set makes sense, since you don't have any plugins that extend the functionality of the QIIME 2 framework. Try installing [`q2-types`](https://github.com/qiime2/q2-types)!

andrewsanchez:
@thermokarst ah ha!  Thank you.  So now the question is, what is the preferred way to display the output of `--show-importable-types`?

It currently looks a bit messy, printing the whole set on one line:
```bash
[aas229@wind ~/projects/qiime_dev/q2cli ]$ qiime tools import --show-importable-types
{DistanceMatrix, FeatureData[PairedEndSequence], Phylogeny[Rooted], SampleData[AlphaDiversity], FeatureTable[PresenceAbsence], Phylogeny[Unrooted], FeatureTable[RelativeFrequency], FeatureData[AlignedSequence], FeatureTable[Frequency], PCoAResults, FeatureData[Sequence], SampleData[PairedEndSequencesWithQuality], FeatureData[Taxonomy], SampleData[SequencesWithQuality]}
```

Would it be better to print each item of the set on it's own line?  If so, is there a better way to do that than:
```python
for name in set_of_importable_types:
    click.echo(name)
```

jairideout:
> Would it be better to print each item of the set on it's own line? If so, is there a better way to do that than:

That strategy sounds good -- can you also sort the types so that they are printed in alphabetical order? (`set` objects aren't ordered)

andrewsanchez:
@jairideout Almost there!   Attempting to sort the set throws a TypeError:

```python
  File "/home/aas229/projects/qiime_dev/q2cli/q2cli/tools.py", line 43, in show_importable_types
    set_of_importable_types = sorted(set_of_importable_types)
TypeError: '<' not supported between instances of '_SemanticType' and '_SemanticType'
```

You can see in the commented lines the various ways I tried to sort the set of importable types, i.e. converting to a list first, etc.

I will continue researching this, but thought I would add another commit just in case the problem is obvious to somebody.

Thanks!

ebolyen:
Hey @andrewsanchez, you are pretty close, but you need one more step:

```python
set_of_importable_types = sorted([repr(t) for t in set_of_importable_types])
```
What this does is convert every type from a `_SemanticType` object, to a `str` (string), which will be sorted alphabetically.

Python allows you to overload operators (like `<`, `>`, `+`, etc...) and the semantic types do overload some operators like `<=` and `>=` to define subtyping relations. But there isn't really a proper *total ordering* for all types, so operators like `>` and `<` aren't implemented which is what `sorted` relies on to figure out how to order things correctly. This is why you would need to call `repr` (or `str`) on every type to get something that can be ordered (using the alphabet).

andrewsanchez:
Thanks everybody for the help!  This is my first pull request, not just for QIIME2 but my first ever in my short history of programming, so any feedback is very welcome.

Would it be appropriate to do an interactive rebase in order to squash these commits into one?  I tried doing that before pushing this latest commit but wasn't able to push.  Not sure what I did wrong, but I can try again.

ebolyen:
@andrewsanchez, when you rewrite history with `git` you need to push with the `-f` (`--force`) flag. This will destructively rewrite the history on the remote. This can make things difficult while review is happening. So when a contributor does this, it's usually at the end of the review process as a final step.

That being said, we've got the "Squash and merge" option set by default for these repositories on GitHub, so a rebase is done automatically. GitHub will convert all commits into a single commit with a message we can edit so you don't even need to worry about it.

andrewsanchez:
@ebolyen That makes sense.  Thanks for the clarification.

jairideout:
@andrewsanchez is this ready for review? Note that there are flake8 errors in the Travis-CI build. You can see the commands that Travis-CI is running by looking in .travis.yml (e.g. to run them locally), and GitHub provides a link to the Travis-CI build output so you can see what was run remotely.

andrewsanchez:
@jairideout Thanks for the tip.  I will keep that in mind in the future.  I corrected the flake8 errors and should be ready for review now.

q2cli/tools.py
```diff
@@ -37,6 +37,21 @@ def export_data(path, output_dir):
     result.export_data(output_dir)
 
 
+def show_importable_types(ctx, param, value):
+    import qiime2.sdk
+
+    PluginManager = qiime2.sdk.PluginManager()
```
jairideout: This is a minor stylistic comment: pep8 recommends using snake case for variables (i.e. lowercase names with words separated by underscores), so a more appropriate name would be something like `plugin_manager`. It's also a little clearer because we're not reusing the class name (`PluginManager`) to represent its instantiation.

That being said, the code can actually be shortened because you don't need a `PluginManager` object to hang around (it's only used once to get the importable types). Here's one way to shorten the code. Also note that I'm using Python's custom [sorting API](https://wiki.python.org/moin/HowTo/Sorting) to avoid taking multiple passes through the `set` (this is faster and more Pythonic).

```python
importable_types = sorted(qiime2.sdk.PluginManager().importable_types, key=repr)
```

You can then loop over this sorted list and print the types with `click.echo`.

q2cli/tools.py
```diff
@@ -37,6 +37,21 @@ def export_data(path, output_dir):
...
+    set_of_importable_types = PluginManager.importable_types
+    set_of_importable_types = [repr(t) for t in set_of_importable_types]
+    set_of_importable_types = sorted(set_of_importable_types)
+
+    if not value or ctx.resilient_parsing:
```
jairideout: Can you move this `if` statement to the very top of the method? (should come before `import qiime2.sdk` because that can be slow). That way we let Click short-circuit the method without doing any extra work.

q2cli/tools.py
```diff
@@ -37,6 +37,21 @@ def export_data(path, output_dir):
...
+        return
+    for name in set_of_importable_types:
```
jairideout: Can you handle the case where there are no importable types? It'd be useful to let users know, printing something like "There are no importable types in the current deployment."

q2cli/tools.py
```diff
@@ -55,6 +70,9 @@ def export_data(path, output_dir):
               help='The format of the data to be imported. If not provided, '
                    'data must be in the format expected by the semantic type '
                    'provided via --type.')
+@click.option('--show-importable-types', is_flag=True, is_eager=True,
```
jairideout: For the `help` string, can you make it a little more descriptive, something like "Show the semantic types that can be supplied to --type to import data into an artifact." ?

jairideout: Can you also expand the `help` string for `--type` to point users to this new option? Something like:

"The semantic type of the artifact that will be created upon importing. Use --show-importable-types to see what importable semantic types are available in the current deployment."

jairideout: Can you also add a unit test for this new option?

andrewsanchez:
Thanks @jairideout 

I'm hoping to find some time to continue working this weekend.

q2cli/tools.py
andrewsanchez: How's this @jairideout ?

```python
    if importable_types:
        for name in importable_types:
            click.echo(name)
    else:
        print('There are no importable types in the current deployment.')
```

q2cli/tools.py
andrewsanchez: @jairideout 

```$ python -m unittest q2cli.tests.test_cli.CliTests``` produces failures for every test.  

Could this be because this branch is behind upstream?  Should I rebase off of upstream?  Or is there a different way you guys run these unit tests?

You can see in my last commit that I started writing a test.   Am I on the right track here?  Is that something like what you're looking for?

File: q2cli/tools.py (hunk 3 above)
Comment (jairideout) [inline q2cli/tools.py]: Looks good! Can you use `click.echo` instead of `print`?

File: q2cli/tools.py (hunk 4 above)
Comment (jairideout) [inline q2cli/tools.py]: > $ python -m unittest q2cli.tests.test_cli.CliTests produces failures for every test.

Use this command to run the unit tests:

```shell
QIIMETEST= nosetests
```

Take a look at `.travis.yml` to see how tests, linting, etc. are run by Travis-CI; you can use those same commands to test locally.

> You can see in my last commit that I started writing a test. Am I on the right track here? Is that something like what you're looking for?

You're on the right track! In addition to checking the exit code, we'd want to assert that some of the expected importable types are in the list (see some of the existing unit tests for how to assert that certain items are in the output).

The q2cli unit tests are written such that they will pass regardless of what plugins you may or may not have installed (e.g. q2-types, q2-feature-table, etc.). When you set the `QIIMETEST` environment variable (see the nosetests command above), that activates a "dummy" plugin defined in the QIIME 2 framework, and avoids loading any other plugins. The dummy plugin defines some importable types (in addition to other things), so you can assert that some of those importable types are in the CLI output.

jairideout:
@andrewsanchez we have a release coming in two weeks and would love to have this included in the release. Do you think you'll have time to finish this PR in the next week?

andrewsanchez:
Definitely!  Sorry for the delay. I should be able to finish it up this weekend now that I'm done with School.

andrewsanchez:
@jairideout  I've given a couple of different solutions here, separated into two different commits.  I'm not sure if it makes sense to do this.  Is this a reasonable way to propose different solutions in a situation like this?

jairideout:
Thanks @andrewsanchez! Separate commits are fine to propose changes, or you could post a code block in a GitHub comment (e.g. similar to what I've been doing in this review).

I prefer your first approach to testing, where you're explicitly verifying that the command produces the expected output. For example:

```python
self.assertTrue('IntSequence1' in result.output)
```

The second approach (where you're looping through `PluginManager.importable_types`) is basically a re-implementation of the CLI command you're testing, so you run the risk of asserting incorrect "expected" output if there's a bug in the implementation itself (it's kind of like circular testing).

q2cli/tools.py
```diff
@@ -37,14 +37,36 @@ def export_data(path, output_dir):
     result.export_data(output_dir)
 
 
+def show_importable_types(ctx, param, value):
+
```
jairideout: Just a quick minor note: we typically don't include a newline at the beginning of the function body unless there's a docstring associated with it.

andrewsanchez: @jairideout Roger on both comments. 

Would you like to `self.assertTrue('IntSequence1' in result.output)` for all expected outputs, as I did in my commit, or is that overkill?  If not, how many exactly do you want to test for?

jairideout: Since there's only a handful of items it probably doesn't hurt to include all of them in the assertions (like you had in that commit). Thanks!

andrewsanchez:
Cool, so any other changes you’d like to see before you merge?  Or is your review for this PR complete?

jairideout:
Other than the unit test updates, can you address this comment?

https://github.com/qiime2/q2cli/pull/131#discussion_r113756942

andrewsanchez:
Whoops, I missed that one!  Let me know if there’s anything else :)

q2cli/tests/test_cli.py
```diff
@@ -66,6 +66,18 @@ def test_plugin_list_commands(self):
         self.assertFalse('split_ints' in commands)
         self.assertFalse('mapping_viz' in commands)
 
+    def test_show_importable_types(self):
+        result = self.runner.invoke(
+            tools, ['import', '--show-importable-types'])
+        self.assertEqual(result.exit_code, 0)
+        self.assertTrue('IntSequence1' in result.output)
```
jairideout: This assertion is repeated below, can you remove the redundant assertion?

andrewsanchez: Got it.  Thanks @jairideout !  This has been a great learning experience  😃  

Improving my knowledge of Git and learning about Travis CI was super valuable!

PR #134, opened 2017-04-26
ENH: Store error on runs without verbose

### LINKED ISSUE #127: save verbose output in a temp file if `--verbose` is not supplied
@ElDeveloper suggested this on the dev call today. Basically do what Sphinx does where stdout/stderr gets logged in a temp directory, unless `--verbose` is supplied. Having this feature would help with getting verbose output if the user doesn't supply `--verbose` and the error takes awhile to trigger.

jakereps:
Changes behavior to write to a temporary file on error, allowing users to retrieve the information without needing to rerun the process using `--verbose`.

Resolves #127

### DIFF
--- a/q2cli/commands.py
+++ b/q2cli/commands.py
@@ -204,17 +204,15 @@ class ActionCommand(click.Command):
         actions_module = importlib.import_module(module_path)
         action = getattr(actions_module, self.action['id'])
 
-        stdout = os.devnull
-        stderr = os.devnull
+        p_read, p_write = os.pipe()
         if verbose:
             # `qiime2.util.redirected_stdio` defaults to stdout/stderr when
             # supplied `None`.
-            stdout = None
-            stderr = None
+            p_write = None
 
         try:
-            with qiime2.util.redirected_stdio(stdout=stdout,
-                                              stderr=stderr):
+            with qiime2.util.redirected_stdio(stdout=p_write,
+                                              stderr=p_write):
                 results = action(**arguments)
         except Exception as e:
             import traceback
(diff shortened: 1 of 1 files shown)


ebolyen:
Do we want to "retroactively capture" stdout and stdin?

jakereps:
Do you mean in addition to the traceback @ebolyen, so like also storing stdout/stdin?

ebolyen:
Yeah! Thats where errors for subprocess calls actually show up.

ebolyen:
@jakereps verbose is actually assuming that stdio hasn't been discarded: 
https://github.com/jakereps/q2cli/blob/4049624333faa4c2b6bb2f02d0a4f20a22c79206/q2cli/commands.py#L207-L218

We should probably put stdio into a temporary file if verbose isn't set. (Technically we just need a file-descriptor, so I think a pipe might also work.) We can then erase that if there isn't an exception.

jairideout:
`s/stdin/stderr/g`

(We don't care about stdin)

q2cli/commands.py
```diff
@@ -217,15 +217,24 @@ def __call__(self, **kwargs):
...
                 click.echo(err=True)
                 self._echo_plugin_error(e, 'See above for debug info.')
             else:
-                self._echo_plugin_error(
-                    e, 'Re-run with --verbose to see debug info.')
+                import tempfile
+                log_path = tempfile.NamedTemporaryFile(prefix='qiime2-err-',
+                                                       suffix='.log',
+                                                       delete=False)
+                with open(log_path.name, 'w+') as fh:
```
jairideout: This won't work on Windows (possibly other platforms) -- it is opening a second file handle to the temp file in `w+` mode (if that file even existed on the filesystem, which differs by platform). `log_path` is already a file-like object in `w+` mode.

I think we want to use a context manager to handle the temporary file object:

```python
with tempfile.NamedTemporaryFile(prefix='qiime2-q2cli-err-', suffix='.log', delete=False) as fh:
    traceback.print_exc(file=fh)
    log_path = fh.name

...
```

Also note that I modified the temp file prefix to include q2cli in the name, just to make it clear what q2 interface is creating the log.

q2cli/commands.py
```diff
@@ -217,15 +217,24 @@ def __call__(self, **kwargs):
...
+                    traceback.print_exc(file=fh)
+                click.echo(err=True)
+                self._echo_plugin_error(e,
+                                        'The full traceback has been saved to '
```
jairideout: Change to:

"Debug info has been saved to `<foo>`."

jakereps:
Thanks @ebolyen and @jairideout! I've addressed both of your comments.

q2cli/commands.py
```diff
@@ -225,16 +223,18 @@ def __call__(self, **kwargs):
...
-                                                       delete=False)
-                with open(log_path.name, 'w+') as fh:
+                with tempfile.NamedTemporaryFile(prefix='qiime2-q2cli-err-',
+                                                 suffix='.log',
+                                                 delete=False, mode='w') as fh:
+                    log_path = fh.name
+                    # Close p_write as .read() searchs for EOF, meaning it will
+                    # hang on the next line that is trying to read the pipe
+                    os.close(p_write)
+                    fh.write(os.fdopen(p_read).read())
```
ebolyen: You may be able to use `os.dup2` here, but I'm not certain.

jairideout:
Thanks @jakereps! I noticed that the pipe's file descriptors can be left open in a few of the codepaths, and unfortunately there isn't a pipe/filedesc context manager so the code may get a bit messy. I have another strategy that can handle this issue and reduce some of the code duplication. Let's discuss when you're in the office today, and if not I'll write up a description here.

PR #136, opened 2017-04-26
ENH: Add --quiet flag to method calls

### LINKED ISSUE #125: Support for --silent
In general, the output information about what artifacts are created is very useful (and I really like clean use of colors). However, there are times when it would be nice to not have this output (e.g., scripting). This could be exposed, similar to the `--verbose` flag as a `--silent` flag, in which output normally directed to stdout is not shown.

jakereps:
Adds `--quiet` flag to `q2cli` that will suppress the success message once a method has been ran and the result objects saved.

Resolves #125

### DIFF
--- a/q2cli/commands.py
+++ b/q2cli/commands.py
@@ -240,9 +240,9 @@ class ActionCommand(click.Command):
                 log.close()
                 os.remove(log.name)
 
-        if not quiet:
-            for result, output in zip(results, outputs):
-                path = result.save(output)
+        for result, output in zip(results, outputs):
+            path = result.save(output)
+            if not quiet:
                 click.secho('Saved %s to: %s' % (result.type, path),
                             fg='green')
 
@@ -266,6 +266,10 @@ class ActionCommand(click.Command):
 
         verbose = self.verbose_handler.get_value(kwargs, fallback=cmd_fallback)
         quiet = self.quiet_handler.get_value(kwargs, fallback=cmd_fallback)
+
+        if verbose and quiet:
+            raise ValueError('Unsure of how to be quiet and verbose at the '
+                             'same time.')
 
         for item in itertools.chain(self.action['signature']['inputs'],
                                     self.action['signature']['parameters']):
--- a/q2cli/commands.py
+++ b/q2cli/commands.py
@@ -133,6 +133,7 @@ class ActionCommand(click.Command):
         self.action = action
         self.generated_handlers = self.build_generated_handlers()
         self.verbose_handler = q2cli.handlers.VerboseHandler()
+        self.quiet_handler = q2cli.handlers.QuietHandler()
         # Meta-Handlers:
         self.output_dir_handler = q2cli.handlers.OutputDirHandler()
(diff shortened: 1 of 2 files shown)


q2cli/commands.py
```diff
@@ -228,9 +230,11 @@ def __call__(self, **kwargs):
                     e, 'Re-run with --verbose to see debug info.')
             click.get_current_context().exit(1)
 
-        for result, output in zip(results, outputs):
-            click.secho('Saved %s to: %s' % (result.type,
-                                             result.save(output)), fg='green')
+        if not quiet:
+            for result, output in zip(results, outputs):
+                path = result.save(output)
```
ebolyen: We still need to save the outputs if it is `--quiet`

jakereps: Ha oops. Wrapped too much.

q2cli/handlers.py
```diff
@@ -135,6 +135,33 @@ def get_value(self, arguments, fallback=None):
...
+    def get_click_options(self):
+        import click
+
+        # `is_flag` will set the default to `False`, but `self._locate_value`
+        # needs to distinguish between the presence or absence of the flag
+        # provided by the user.
+        yield click.Option(
+            ['--' + self.cli_name], is_flag=True, default=None,
+            help='Silence output to stdout and/or stderr during '
+                 'execution of this action.  [default: %s]' % self.default)
```
ebolyen: Discussed this with @jairideout, we'd like to see the help text look like:

```
Silence output if execution is successful (silence is golden).  [default: <whatever>]
```

q2cli/commands.py
```diff
@@ -184,7 +186,7 @@ def __call__(self, **kwargs):
         import os
         import qiime2.util
 
-        arguments, missing_in, verbose = self.handle_in_params(kwargs)
+        arguments, missing_in, verbose, quiet = self.handle_in_params(kwargs)
```
ebolyen: Raise an exception if quiet and verbose are both true. (We're planning to make these mutually exclusive)

jakereps:
Thanks @ebolyen! I've addressed your comments.

q2cli/commands.py
```diff
@@ -251,6 +255,11 @@ def handle_in_params(self, kwargs):
         cmd_fallback = self.cmd_config_handler.get_value(kwargs)
 
         verbose = self.verbose_handler.get_value(kwargs, fallback=cmd_fallback)
+        quiet = self.quiet_handler.get_value(kwargs, fallback=cmd_fallback)
+
+        if verbose and quiet:
+            raise ValueError('Unsure of how to be quiet and verbose at the '
```
jairideout: Can you use the standard way of error reporting in Click? i.e. `click.secho(err=True, fg='red', bold=True)` and `click.exit(1)`?

We really need a helper for this -- I noticed we have a few different ways of error handling in q2cli. But that's for another day...

jakereps:
Thanks @jairideout! I've swapped the method used in creating the quiet/verbose error.

PR #137, opened 2017-05-05
WIP: Metadata artifact

mortonjt:
This is another bit pair-programmed with @ebolyen .  This basically makes the CLI aware of the metadata artifact

### DIFF
--- a/q2cli/handlers.py
+++ b/q2cli/handlers.py
@@ -310,23 +310,20 @@ class MetadataHandler(Handler):
 
     def get_value(self, arguments, fallback=None):
         import qiime2
-        import qiime2.sdk.util
-        path = self._locate_value(arguments, fallback)
 
+        path = self._locate_value(arguments, fallback)
         if path is None:
             return None
-
         try:
             # check to see if path is an artifact
-            art = qiime2.Artifact.load(path)
+            artifact = qiime2.Artifact.load(path)
         except Exception:
             pass
         else:
-            if qiime2.sdk.util.has_metadata(art):
-                return art.view(qiime2.Metadata)
+            if artifact.has_metadata():
+                return qiime2.Metadata.from_artifact(artifact)
             else:
-                raise ValueError("Cannot load Artifact %r into Metadata"
-                                 ". No transformer found" % path)
+                raise ValueError("Artifact (%s) has no metadata." % path)
 
         return qiime2.Metadata.load(path)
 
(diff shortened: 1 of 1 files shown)


q2cli/handlers.py
```diff
@@ -310,12 +310,25 @@ def get_click_options(self):
...
-
+        import qiime2.sdk.util
         path = self._locate_value(arguments, fallback)
+
         if path is None:
             return None
+
+        try:
+            # check to see if path is an artifact
+            art = qiime2.Artifact.load(path)
```
ebolyen: @jairideout we couldn't come up with a better way to check this. Should we create an API, or is this fine for now?

q2cli/handlers.py
```diff
@@ -310,12 +310,25 @@ def get_click_options(self):
...
+        except Exception:
+            pass
         else:
-            return qiime2.Metadata.load(path)
+            if qiime2.sdk.util.has_metadata(art):
+                return art.view(qiime2.Metadata)
```
ebolyen: This should be `qiime2.Metadata.from_artifact(art)` to capture the provenance. (This is the one we undo'ed a bunch, looks like it may not have been redo'ed)

q2cli/handlers.py
```diff
@@ -377,8 +390,20 @@ def get_value(self, arguments, fallback=None):
...
+        path, category = values
+        try:
+            # check to see if path is an artifact
+            art = qiime2.Artifact.load(path)
+        except Exception:
+            pass
         else:
-            return qiime2.MetadataCategory.load(*values)
+            if qiime2.sdk.util.has_metadata(art):
+                return art.view(qiime2.Metadata).get_category(category)
```
ebolyen: Same here, `MetadataCategory.from_artifact(art, category)`

q2cli/handlers.py
```diff
@@ -377,8 +390,20 @@ def get_value(self, arguments, fallback=None):
...
+            else:
+                raise ValueError("Cannot load Artifact %r into Metadata"
+                                 ". No transformer found" % path)
```
ebolyen: `Artifact has no metadata.`

q2cli/handlers.py
```diff
@@ -310,12 +310,25 @@ def get_click_options(self):
 
     def get_value(self, arguments, fallback=None):
         import qiime2
-
+        import qiime2.sdk.util
```
ebolyen: it looks like the `sdk.util` subpackage gets loaded when you import `qiime2`, so maybe we can skip this import.

q2cli/handlers.py
jairideout: Ideally we'd have an API to check whether it's a valid QIIME 2 artifact or visualization (I created an issue [here](https://github.com/qiime2/qiime2/issues/259)) but this is fine for now. It'd be handy to have a TODO comment linking to the issue I created.

q2cli/handlers.py
```diff
@@ -310,12 +310,25 @@ def get_click_options(self):
...
             return None
+
+        try:
+            # check to see if path is an artifact
+            art = qiime2.Artifact.load(path)
+        except Exception:
+            pass
         else:
-            return qiime2.Metadata.load(path)
+            if qiime2.sdk.util.has_metadata(art):
```
jairideout: How about moving `.has_metadata()` to a method on the `Artifact` class? It'd make it easier to discover instead of being orphaned in `sdk.util`.

jairideout: Sorry, just noticed you had pinged me on that [here](https://github.com/qiime2/qiime2/pull/249#discussion_r115625672)

PR #144, opened 2017-06-12
BUG: disallow repeated options

### LINKED ISSUE #143: CLI permits repeated options
Click permits options to be repeated on the CLI and will always use the value of the *last* specified option (this is [intentional, documented behavior](http://click.pocoo.org/6/options/#multiple-options)). I can't find a way to disallow repeated options in Click but it should be possible to hack something together.

It seems like we'd want q2cli to distinguish between single and repeated options, especially once metadata merging (#140) is supported. Ideally we'll be able to make the CLI error on repeated options *unless* the option is metadata.

jairideout:
Click allows options to be repeated on the command line and stores the value of the last specified option (this is to support overriding options set in shell aliases). While this is common behavior in CLI tools, it is prevented in q2cli to avoid confusion with options that are intended to be supplied multiple times (``multiple=True``; Click calls these "multiple options"). QIIME 2 metadata is an example of a "multiple option" in q2cli.

To support this behavior, q2cli now has ``q2cli.Option`` and ``@q2cli.option``, which are intended to be drop-in replacements for their Click counterparts.

Fixes #143.

### DIFF
--- a/q2cli/core.py
+++ b/q2cli/core.py
@@ -1,0 +1,108 @@
+# ----------------------------------------------------------------------------
+# Copyright (c) 2016-2017, QIIME 2 development team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+
+import click
+
+import q2cli.util
+
+
+class Option(click.Option):
+    """``click.Option`` with customized behavior for q2cli.
+
+    Note to q2cli developers: you'll generally want to use this class and its
+    corresponding decorator (``@q2cli.option``) over ``click.Option`` and
+    ``@click.option`` to keep a consistent CLI behavior across commands. This
+    class and decorator are designed to be drop-in replacements for their Click
+    counterparts.
+
+    """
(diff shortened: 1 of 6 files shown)


q2cli/core.py
```diff
@@ -0,0 +1,108 @@
...
+#
+# The full license is in the file LICENSE, distributed with this software.
+# ----------------------------------------------------------------------------
+
+import click
+
+import q2cli.util
+
+
+class Option(click.Option):
```
thermokarst: 👍 on naming (I left as @jairideout & @ebolyen started chatting about this IRL)

q2cli/core.py
```diff
@@ -0,0 +1,108 @@
...
+        are intended to be supplied multiple times (``multiple=True``; Click
+        calls these "multiple options"). QIIME 2 metadata is an example of a
+        "multiple option" in q2cli.
+
+        References
+        ----------
+        .. [1] http://click.pocoo.org/6/options/#multiple-options
+
+        """
+        # General strategy:
```
thermokarst: Thanks for the roadmap --- this'll be handy!

PR #148, opened 2017-07-12
MAINT: clarify help text

gregcaporaso:
We should transition from calling these "metadata mapping files" to "metadata files", as the term "mapping" doesn't really get us much.


jairideout:
Thanks @gregcaporaso!

Note: I originally wrote this text to match what we have in the [metadata tutorial](https://docs.qiime2.org/2017.6/tutorials/metadata/), it's called a "sample/feature metadata mapping file" there. We can update the tutorial's text before the release since we'll be making some significant edits to the doc anyways.

PR #155, opened 2017-08-22
ENH: let `make test` handle QIIMETEST env var

jairideout:
Previously:

```
QIIMETEST= make test
```

Now:

```
make test
```

:tada:

Also removed the `test-cov` target because coverage isn't set up for this package (yet).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,6 @@ install:
   - make install
 script:
   - make lint
-  - QIIMETEST= make test-cov
+  - make test
   - QIIMETEST= source tab-qiime
-  - QIIMETEST= qiime
+  - QIIMETEST= qiime info
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all lint test test-cov install dev clean distclean
+.PHONY: all lint test install dev clean distclean
 
 all: ;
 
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -19,6 +19,6 @@ install:
   - make install
 script:
   - make lint
-  - QIIMETEST= make test-cov
+  - make test
   - QIIMETEST= source tab-qiime
-  - QIIMETEST= qiime
```
jairideout: Running `qiime info` instead of `qiime` has the opportunity to fail b/c it does more work, plus the command gives some useful debug info in the Travis-CI log.

Makefile
```diff
@@ -7,10 +7,7 @@ lint:
 	flake8
 
 test: all
-	nosetests
-
-test-cov: all
```
thermokarst: `test-cov` was one of the entrypoints we had agreed would be in all `Makefiles`. I am good with removing it (especially since we don't run coverage checks on this repo), but I thought it worth mentioning.

thermokarst:
I was waffling back and forth on setting up env var pass-through (or at least adding a TODO to the Makefile), but it seems like something we won't miss in the future, and it buys us next-to nothing right now.

PR #166, opened 2017-10-20
ENH: Add `Set` support

### LINKED ISSUE #159: Support `Set` primitive
This would be really useful for visualizations like `diversity alpha-rarefaction`, or methods like `diversity alpha` and `diversity beta`

jakereps:
Resolves #159

### DIFF
--- a/q2cli/handlers.py
+++ b/q2cli/handlers.py
@@ -7,7 +7,6 @@
 # ----------------------------------------------------------------------------
 
 import collections
-import json
 
 # Sentinel to avoid the situation where `None` *is* the default value.
 NoDefault = collections.namedtuple('NoDefault', [])()
@@ -559,7 +558,9 @@ class RegularParameterHandler(GeneratedHandler):
             if type(value) is not bool:
                 value = self._parse_boolean(value)
             return value
+        elif self.ast['name'] == 'Set':
+            return set(value)
         else:
             import qiime2.sdk
             return qiime2.sdk.parse_type(
-                self.repr, expect='primitive').decode(json.dumps(value))
+                self.repr, expect='primitive').decode(value)
--- a/q2cli/handlers.py
+++ b/q2cli/handlers.py
@@ -520,6 +520,9 @@ class RegularParameterHandler(GeneratedHandler):
                                   help='[default: %s]' % self.default,
                                   multiple=multiple)
 
+        if multiple:
+            option.help = 'May be supplied multiple times.  ' + option.help
+
         yield self._add_description(option)
 
     def get_value(self, arguments, fallback=None):
@@ -534,6 +537,8 @@ class RegularParameterHandler(GeneratedHandler):
                 value = self._parse_boolean(value)
             return value
         elif self.ast['name'] == 'Set':
+            if len(value) > len(set(value)):
+                self._error_with_duplicate_in_set(value)
             return set(value)
         elif self.ast['name'] == 'List':
             return list(value)
@@ -541,3 +546,17 @@ class RegularParameterHandler(GeneratedHandler):
             import qiime2.sdk
             return qiime2.sdk.parse_type(
                 self.repr, expect='primitive').decode(value)
+
+    def _error_with_duplicate_in_set(self, elements):
+        import click
+        import collections
+
+        counter = collections.Counter(elements)
+        dups = {name for name, count in counter.items() if count > 1}
+
+        ctx = click.get_current_context()
+        click.echo(ctx.get_usage() + '\n', err=True)
+        click.secho("Error: Option --%s was given these values: %r more than "
+                    "one time, values passed should be unique."
+                    % (self.cli_name, dups), err=True, fg='red', bold=True)
+        ctx.exit(1)
--- a/q2cli/handlers.py
+++ b/q2cli/handlers.py
@@ -7,6 +7,7 @@
 # ----------------------------------------------------------------------------
 
 import collections
+import json
 
 # Sentinel to avoid the situation where `None` *is* the default value.
 NoDefault = collections.namedtuple('NoDefault', [])()
@@ -561,4 +562,4 @@ class RegularParameterHandler(GeneratedHandler):
         else:
             import qiime2.sdk
             return qiime2.sdk.parse_type(
-                self.repr, expect='primitive').decode(value)
+                self.repr, expect='primitive').decode(json.dumps(value))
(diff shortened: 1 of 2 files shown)


q2cli/handlers.py
```diff
@@ -553,6 +558,8 @@ def get_value(self, arguments, fallback=None):
             if type(value) is not bool:
                 value = self._parse_boolean(value)
             return value
+        elif self.ast['name'] == 'Set':
+            return set(value)
```
ebolyen: `Set` is a generic and has a parameterized type: `Set[Str % Choices({'foo', 'bar'})]`, `Set[Int]`, etc. So we'd probably need to dispatch the handler again for every element in `value` on the contained type (`Str % Choices({'foo', 'bar'})`, `Int`).

jakereps: Testing it with the `q2-diversity` PR that uses `Set[Str % Choices(...)]` worked as expected. I hadn't gotten a chance to write any actual tests for it though.

jakereps:
```bash
(qiime2) caporasolab $ qiime diversity alpha-rarefaction --i-table
~/Desktop/qiime2-moving-pictures-tutorial/table.qza --i-phylogeny
~/Desktop/qiime2-moving-pictures-tutorial/rooted-tree.qza --p-max-depth 10000
--o-visualization test --p-metric faith_pd --p-metric pielou_e
Saved Visualization to: test.qzv
(qiime2) caporasolab $ qiime diversity alpha-rarefaction --i-table
~/Desktop/qiime2-moving-pictures-tutorial/table.qza --i-phylogeny
~/Desktop/qiime2-moving-pictures-tutorial/rooted-tree.qza --p-max-depth 10000
--o-visualization test --p-metric faith_pd --p-metric pielou_e --p-metric error
Plugin error from diversity:

  Argument to parameter 'metric' is not a subtype of Set[Str %
  Choices({'ace', 'berger_parker_d', 'brillouin_d', 'chao1',
  'dominance', 'doubles', 'enspie', 'faith_pd', 'fisher_alpha',
  'gini_index', 'goods_coverage', 'heip_e', 'lladser_pe', 'margalef',
  'mcintosh_d', 'mcintosh_e', 'menhinick', 'michaelis_menten_fit',
  'observed_otus', 'pielou_e', 'robbins', 'shannon', 'simpson',
  'simpson_e', 'singles'})].

Debug info has been saved to /var/folders/zc/30qh4_v57slc7nh0g9xqw7lh0000gn/T/qiime2-q2cli-err-nxeqen55.log
(qiime2) caporasolab $ 
```

jairideout: :+1: 

I bet we can get Click to give a clearer error message if the option is declared as a "multi-option" (`multiple=True`) of `click.Choices`. Would you mind testing that out?

jakereps: Yeah I noticed it doesn't clarify. I can dig into it!

ebolyen: I suspect you will need to kind of refactor the way `RegularParameterHandler` works (this current implementation has always been a hack). You'll need some kind of way to build up the click option that should be returned by walking through the primitive type. (And then do the same thing when getting a value, so that the following test works:

User passes
```bash
--p-some-set-input 1 --p-some-set-input 2 --p-some-set-input 3
```
Where  the type is: 
```
some_set_input: Set[Int]
```
So that q2cli returns `{1, 2, 3}` and not `{"1", "2", "3"}` which is what I _think_ the current code does, since it and click are unaware of the `Int` inside of `Set`.

jakereps: Oh ok. I see what you mean now. QIIME 2 was correctly handling choices, etc but click will cast to string no matter what at the moment.

jakereps:
Thanks @ebolyen and @jairideout! I think everything has been addressed.

q2cli/handlers.py
```diff
@@ -504,6 +504,11 @@ def get_type(self):
                 return mapping['Float']
             else:
                 raise NotImplementedError()
+        # Specify 'Set'/'List' because 'Dict' will require more modifications
+        elif self.ast['name'] in ['Set', 'List']:
+            # n-`primitive`s can not be unioned, so ignore trailing values
+            field, *_ = self.ast['fields']
+            return mapping[field['name']]
```
ebolyen: This should handle the same predicate logic as above. So ideally we would return a `click.Choice` or `click.IntRange` if appropriate.

q2cli/handlers.py
```diff
@@ -504,6 +504,11 @@ def get_type(self):
                 return mapping['Float']
             else:
                 raise NotImplementedError()
+        # Specify 'Set'/'List' because 'Dict' will require more modifications
+        elif self.ast['name'] in ['Set', 'List']:
+            # n-`primitive`s can not be unioned, so ignore trailing values
+            field, *_ = self.ast['fields']
```
ebolyen: This is guaranteed to be a single field, so you can just unpack the first value:
```python
field, = self.ast['fields']
```

q2cli/handlers.py
ebolyen: I would create a helper `_convert_primitive` which does the mess of logic that `get_type` usually does, then you can make `get_type` just check:

```python
if self.ast['name'] in 'Set', 'List':
    field, = self.ast['fields']
    return _convert_primitive(field)
else:
    return _convert_primitive(self.ast)
```

ebolyen:
Sorry one more requirement, we should check for duplicates in the value before coercing to a `set`, that way it kind of matches how single-options work, where supplying duplicate values isn't allowed.

Also, we should handle `List` in `get_value` since the logic is basically the same (sans duplicate checking) and `get_type` already understands it.

jakereps:
Thanks @ebolyen, should be ready for a final pass! 👍

ebolyen:
@jakereps, I made a squash commit on your branch with a couple extra things (help text + duplicate checking). Let me know if that looks good to you.

P.S. we are a little stuck test-wise because the framework doesn't have any actions that use sets right now.

jakereps:
Yeah I was poking around trying to find a way to write a test, but without also adding something to the dummy plugin it didn't seem like a quick thing to get up. The logic changes all look good to me, only aesthetic thing would be the help text. Should the note be added at the end in parens? It kind of gets buried at the moment:

image

ebolyen:
@jakereps, I was matching what our other more complicated options do, they seem to put the instructions before the `[optional]`

ebolyen:
Thanks @jakereps!

PR #176, opened 2018-01-26
BUG: preserve Metadata ID header when not merging Metadata

jairideout:
Previously, when a single Metadata was provided as input, the ID header was
normalized to 'id' instead of keeping the original value (e.g. '#SampleID').

New behavior is to retain original ID header when not merging, and normalize
the ID header to 'id' when a merge happens.

Depends on https://github.com/qiime2/qiime2/pull/367


q2cli/tests/test_cli.py
```diff
@@ -433,13 +438,35 @@ def test_single_metadata(self):
                 result, exp_tsv=exp_tsv,
                 exp_yaml="metadata: !metadata 'metadata.tsv'")
 
+    def test_single_metadata_alt_id_header(self):
```
thermokarst: 👍

PR #178, opened 2018-02-13
ENH: adds `inspect-metadata` and cleans up errors

### LINKED ISSUE #165: Passing in a visualization gives cryptic error
This [came up on the forum](https://forum.qiime2.org/t/getting-error-when-performing-diversity-core-metrics-phylogenetic/1446). It looks like we're getting an error because [this line](https://github.com/qiime2/q2cli/blob/2273ccce1a8934cb41b1269d855207e6b1948721/q2cli/handlers.py#L302) errors without knowing the context. So a user who passes in a `qzv` will see something about using `Visualization.load` instead.

### LINKED ISSUE #177: Add a tool command for viewing metadata columns/summaries
Originally from https://github.com/qiime2/qiime2/issues/363

Name would be something like `qiime tools inspect-metadata` or `qiime tools summarize-metadata` and would provide a list of the columns and their type annotations/inference (and in the future, whatever other kinds of properties we associate with column names).

ebolyen:
fixes #177
fixes #165 

---

image

image

image

image

image

---

image

### DIFF
--- a/q2cli/tools.py
+++ b/q2cli/tools.py
@@ -195,9 +195,11 @@ def inspect_metadata(paths, tsv, failure):
     if not tsv:
         click.secho(formatter("=" * max_name_len, "=" * max_type_len),
                     bold=True)
-        click.secho(("{0:>%d}  " % max_name_len).format("IDS:"), bold=True, nl=False)
+        click.secho(("{0:>%d}  " % max_name_len).format("IDS:"),
+                    bold=True, nl=False)
         click.echo(metadata.id_count)
-        click.secho(("{0:>%d}  " % max_name_len).format("COLUMNS:"), bold=True, nl=False)
+        click.secho(("{0:>%d}  " % max_name_len).format("COLUMNS:"),
+                    bold=True, nl=False)
         click.echo(metadata.column_count)
 
 
--- a/q2cli/tools.py
+++ b/q2cli/tools.py
@@ -144,27 +144,21 @@ def peek(path):
 
 
 @tools.command('inspect-metadata',
-               short_help='Inpect columns available in metadata.')
-@q2cli.option('--tsv/--no-tsv', default=False)
-@click.argument('path', type=click.Path(exists=True, file_okay=True,
-                                        dir_okay=False, readable=True))
+               short_help='Inpect columns available in metadata.',
+               help='Inspect metadata files or artifacts viewable as metadata.'
+                    ' Providing multiple file paths to this command will merge'
+                    ' the metadata.')
+@q2cli.option('--tsv/--no-tsv', default=False,
+              help='Print as machine-readable TSV instead of text.')
+@click.argument('paths', nargs=-1, required=True,
+                type=click.Path(exists=True, file_okay=True, dir_okay=False,
+                                readable=True))
 @q2cli.util.pretty_failure(traceback=None)
-def inspect_metadata(path, tsv, failure):
-    import qiime2
-    import qiime2.sdk
-
-    try:
-        artifact = qiime2.sdk.Result.load(path)
-    except Exception:
-        metadata = qiime2.Metadata.load(path)
-    else:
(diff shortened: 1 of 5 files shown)


q2cli/tools.py
```diff
@@ -145,6 +143,62 @@ def peek(path):
...
+    import qiime2
+    import qiime2.sdk
+
+    try:
+        artifact = qiime2.sdk.Result.load(path)
+    except Exception:
+        metadata = qiime2.Metadata.load(path)
+    else:
+        if isinstance(artifact, qiime2.Visualization):
+            raise Exception("Visualizations do not have metadata.")
```
jairideout: Can you update this text to be consistent with the error message used in the MetadataHandler / MetadataColumnHandler when something can't be viewed as metadata? That'll also match the language in the `docs` and framework's error messages.

"Visualizations cannot be viewed as QIIME 2 Metadata."

q2cli/tools.py
```diff
@@ -145,6 +143,62 @@ def peek(path):
...
+        elif artifact.has_metadata():
+            metadata = artifact.view(qiime2.Metadata)
+        else:
+            raise Exception("Artifact of %r does not have metadata."
```
jairideout: similar comment as above:

"Artifact %r cannot be viewed as QIIME 2 Metadata."

I think using the Artifact's repr (instead of its type) gives the user a little more info because it includes the UUID and type.

ebolyen: I tried that initially and thought it looked a little strange seeing its python repr "bleed through" the command line

ebolyen: with variadic input, I think I'll include the path and UUID in prose since it's no longer clear which one is responsible for the issue

ebolyen:
I like including the metadata dimensions, I would probably put it at the top and exclude it from the `--tsv` results. Does that sounds like a good plan?

jairideout:
> I like including the metadata dimensions, I would probably put it at the top and exclude it from the --tsv results. Does that sounds like a good plan?

:+1:

q2cli/tools.py
jairideout: :+1: , that's a similar approach as MetadataHandler

ebolyen:
Turns out it looks better at the bottom, and make it intuitively look like a summary total (which it is!)

image

ebolyen:
Updated errors:
image

q2cli/util.py
```diff
@@ -31,31 +31,61 @@ def to_cli_name(name):
...
+                 traceback='stderr', status=1):
+        self.header = header
+        self.traceback = traceback
+        self.status = status
+
+    def __call__(self, function):
+        def wrapped(*args, **kwargs):
+            with self:
+                return function(*args, **kwargs, failure=self)
+        # not using functools.wraps to keep import overhead low
```
thermokarst: 👍

q2cli/tools.py
```diff
@@ -145,6 +143,88 @@ def peek(path):
...
+        click.secho(("{0:>%d}  " % max_name_len).format("COLUMNS:"),
+                    bold=True, nl=False)
+        click.echo(metadata.column_count)
+
+
+def _load_metadata(path):
+    import qiime2
+    import qiime2.sdk
+
+    # TODO: clean up duplication between this and the metadata handlers.
```
thermokarst: 👍

q2cli/tools.py
```diff
@@ -145,6 +143,88 @@ def peek(path):
...
+        import io
+
+        def formatter(*row):
+            # This is gross, but less gross than robust TSV writing.
+            with io.StringIO() as fh:
+                writer = csv.writer(fh, dialect='excel-tab', lineterminator='')
+                writer.writerow(row)
+                return fh.getvalue()
+    else:
+        formatter = ("{0:>%d}  {1:%d}" % (max_name_len, max_type_len)).format
```
thermokarst: You're in the mainframe now!

PR #179, opened 2018-02-14
MAINT: update unit test to match Metadata writer number formatting

jairideout:
Depends on https://github.com/qiime2/qiime2/pull/371


thermokarst:
Merging now even though travis ci is red - busywork will sort itself out.

PR #183, opened 2018-04-17
BUG: --verbose now shows stack trace again

ebolyen:


ebolyen:
oops, thanks @thermokarst!

PR #189, opened 2018-06-18
ENH: Export file with desired output format

turanoo:


q2cli/tests/test_tools.py
```diff
@@ -130,6 +135,68 @@ def test_merged_metadata(self):
...
+
+    def test_export_to_dir_no_format(self):
+        output_path = os.path.join(self.tempdir, 'output')
+        self.runner.invoke(tools, [
+            'export', '--input-path', self.viz, '--output-path', output_path
+        ])
+
+        self.assertTrue(os.path.isdir(output_path))
+        self.assertIn('index.html', os.listdir(output_path))
+        self.assertIn('index.html', os.listdir(output_path))
```
ebolyen: Drop duplicate assert

q2cli/tests/test_tools.py
```diff
@@ -130,6 +135,68 @@ def test_merged_metadata(self):
...
+
+    def test_export_to_file(self):
+        output_path = os.path.join(self.tempdir, 'output')
+        os.mkdir(output_path)
+        self.runner.invoke(tools, [
+            'export', '--input-path', self.ints1, '--output-path', output_path,
+            '--output-format', 'IntSequence1'
+            ])
+        print(os.listdir(output_path))
```
ebolyen: Looks like this print never became an assert

q2cli/tests/test_tools.py
```diff
@@ -130,6 +135,68 @@ def test_merged_metadata(self):
...
+    #
+    #     self.runner.invoke(tools, [
+    #         'export', '--input-path', self.ints1, '--output-path', output_path,
+    #         '--output-format', 'IntSequence1'
+    #     ])
+    #     self.assertTrue(os.path.isfile(output_path))
+    #
+    #
+    # def test_export_path_dir_is_replaced(self):
+    #     assert False
```
ebolyen: Tests should never be commented out. Please finish these.

q2cli/tests/test_tools.py
```diff
@@ -130,6 +135,64 @@ def test_merged_metadata(self):
...
+
+    def test_export_path_file_is_replaced(self):
+        output_path = os.path.join(self.tempdir, 'helloWorld')
+        os.mkdir(output_path)
+
+        result = self.runner.invoke(tools, [
+            'export', '--input-path', self.ints1, '--output-path', output_path,
+            '--output-format', 'IntSequence1'
+        ])
+        self.assertNotIn('helloWorld', result.output)
```
ebolyen: You'll want to check that it isn't in `output_path` rather than the stdout.

q2cli/tests/test_tools.py
```diff
@@ -130,6 +135,64 @@ def test_merged_metadata(self):
...
+
```
ebolyen: I was going to comment that we're missing a test to see if a directory is replaced, but in retrospect, that probably doesn't make a lot of sense. No reason to delete the directory before-hand, that could cause a lot of unhappy users.

q2cli/tools.py
```diff
@@ -19,26 +19,35 @@ def tools():
...
+@q2cli.option('--input-path', required=True,
+              type=click.Path(exists=True, file_okay=True,
+                              dir_okay=False, readable=True))
+@q2cli.option('--output-path', required=True,
+              type=click.Path(exists=False, file_okay=True, dir_okay=True,
                               writable=True),
               help='Directory where data should be exported to')
-def export_data(path, output_dir):
+@q2cli.option('--output-format', required=False,
+              help='Format which the data should be exported as')
```
ebolyen: Mention that this cannot be used with a visualization.

q2cli/tools.py
```diff
@@ -19,26 +19,35 @@ def tools():
...
-                    "Visualization and will support exporting to multiple "
-                    "formats in the future. For now, the data is exported in "
-                    "the format it is stored within the Artifact or "
-                    "Visualization. Use 'qiime tools extract' to extract the "
-                    "Artifact or Visualization's entire archive.")
-@click.argument('path', type=click.Path(exists=True, file_okay=True,
-                                        dir_okay=False, readable=True))
-@q2cli.option('--output-dir', required=True,
-              type=click.Path(exists=False, file_okay=False, dir_okay=True,
+               short_help='Export data from a QIIME 2 Artifact',
```
ebolyen: append: `or Visualization.`.

I think that got lost in the reflow

q2cli/tools.py
```diff
@@ -19,26 +19,35 @@ def tools():
...
+               help="Exporting extracts the data stored in an Artifact"
```
ebolyen: Try:
`"Exporting extracts (and optionally transforms) data stored inside an Artifact or Visualization. Note that Visualizations cannot be transformed with --output-format."` reflow as needed for PEP8.

ebolyen:
Looks like one of the tests is failing as well.

q2cli/tools.py
turanoo: @ebolyen, thanks for reviewing the PR, I will go ahead and make these changes!

turanoo:
@ebolyen I will have a look at that!

q2cli/tests/test_tools.py
```diff
@@ -130,6 +135,70 @@ def test_merged_metadata(self):
...
+    def test_export_path_file_is_replaced(self):
+        output_path = os.path.join(self.tempdir, 'helloWorld')
+        os.mkdir(output_path)
+        self.runner.invoke(tools, [
+            'export', '--input-path', self.ints1, '--output-path', output_path,
+            '--output-format', 'IntSequence1'
+        ])
+
+        self.assertNotIn('helloWorld', os.listdir(output_path))
```
ebolyen: This isn't the right kind of assertion, remember, we're looking to see that the file is replaced.
The rough idea here is right, but we need to look inside a file not a directory.

turanoo: Yeah, you're right! Thanks for the reminding me.

ebolyen:
@thermokarst, what are you thoughts on merging this so close to the release? We'll probably run into documentation updates since we're breaking the API. It might also be nice to break the API on `import` to match (`--source-format` -> `--input-format`).

ebolyen:
Discussed offline, queued for next release cycle :factory:

PR #196, opened 2018-08-29
NEW: automatic tab completion on env activation

ebolyen:
Adds automatic tab completion when using ZSH or bash after activating the conda environment.


ElDeveloper:
source

PR #197, opened 2018-10-01
TST: Update tests for click 7.0

thermokarst:


thermokarst:
This change should pass with both click 6.7 and 7.0.

PR #199, opened 2018-11-08
MAINT: Updating brittle test

thermokarst:


thermokarst:
Travis will fail because the framework is out of sync. Tested locally. This is related to the changes introduced in https://github.com/qiime2/qiime2/pull/417

PR #204, opened 2019-01-17
BUG: fixes export for EXDEV situations

ebolyen:
This should correct issues with cross-device links (e.g. os.rename).

I don't have an obvious way to test this, but I could try something with mocks maybe.
TODO:
 - [x] Write a test or two

### DIFF
--- a/q2cli/tests/test_tools.py
+++ b/q2cli/tests/test_tools.py
@@ -7,10 +7,8 @@
 # ----------------------------------------------------------------------------
 
 import os
-import errno
 import shutil
 import unittest
-import unittest.mock as mock
 import tempfile
 
 from click.testing import CliRunner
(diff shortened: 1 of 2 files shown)


ebolyen:
So it turns out mocks don't really help here as I can only mock the `os.renames` which is implementation specific. And there's not a good way to scaffold an entire filesystem for a test.

I would really need to mock the syscall which is OS specific so I'm giving up on testing that EXDEV is working correctly. It should be fine since `util.duplicate` solves EXDEV specifically and that seems to be working fine in the wild.

thermokarst:
> And there's not a good way to scaffold an entire filesystem for a test

/gif great scott.

ebolyen:
Thanks @turanoo!

thermokarst:
Thanks @ebolyen - can you please update the changelog when you get a chance?

PR #206, opened 2019-01-25
BUG: ignore unset var in conditional activation

### LINKED ISSUE #200: bash unbound variable bug (conda hook)
**Bug Description**
> (text copied verbatim from originating forum issue (Reference 1, below).

When activating a qiime2 2018.11 conda environment from within a bash script that has the `-u` option activated, q2cli's tab completion hook raises the following error message:

```shell
$HOME/miniconda3/envs/qiime2-2018.11/etc/conda/activate.d/50_activate_q2cli_tab_completion.sh: line 1: ZSH_VERSION: unbound variable
```

```bash
#!/usr/bin/env bash
set -u

source activate qiime2-2018.11
```

The error message doesn't occur if `-u` is turned off before activating the conda environment:

```bash
#!/usr/bin/env bash
set -u

set +u
source activate qiime2-2018.11
set -u
```

**References**
1. [Original forum report](https://forum.qiime2.org/t/q2cli-zsh-version-unbound-variable-error/7516)

ebolyen:
fixes #200 

See: https://www.gnu.org/software/bash/manual/bash.html#Shell-Parameter-Expansion

Unfortunately, it looks like there are some other scripts with this problem, namely:
https://github.com/bioconda/bioconda-recipes/blob/c2e55df081bd1e8449bc64949d2d3f6d1a080f63/recipes/arb-bio/activate.sh

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,9 @@ install: all
 	cp hooks/50_activate_q2cli_tab_completion.sh $(PREFIX)/etc/conda/activate.d/
 
 dev: all
-	pip install -e .
+	pip install -e . && \
+	mkdir -p $(PREFIX)/etc/conda/activate.d && \
+	cp hooks/50_activate_q2cli_tab_completion.sh $(PREFIX)/etc/conda/activate.d/
 
 clean: distclean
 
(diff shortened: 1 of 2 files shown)


ebolyen:
x-ref https://github.com/bioconda/bioconda-recipes/pull/13231

PR #209, opened 2019-01-30
Also show package in plugin --version (fixes #207)

### LINKED ISSUE #207: Misleading output in `qiime pluginname --version`
e.g.
```
$ qiime vsearch --version
vsearch version 2019.1.0
```

This looks a lot like `vsearch` version `2019.1.0` is installed - which doesn't exist. It *might* even be interpreted as `q2-vsearch` as of version `2019.1.0` is installed. But that is also wrong. The command really prints the version of a `vsearch` *plugin* that probably has been registered by `q2-vsearch` and hopefully has the same version. But not necessarily so...

Since a single plugin *package* can provide multiple *Plugin* objects, and each can have separate versions, a precise version string would be something like 

```
QIIME 2 Plugin 'vsearch' version 2019.1.0 (from package 'q2-vsearch' version 2019.1.0)
```

(as unfortunate as that looks - but `q2-vsearch` could have version `1.2.3` while registering the plugins `align` and `make_coffee` with versions `dev15` and `3.2.1.2019`, respectively.)

Just showing the plugin *name* raw, given the custom of using the name of the tool as the name of the plugin, is misleading, though.

epruesse:
fixes #207

### DIFF
--- a/q2cli/commands.py
+++ b/q2cli/commands.py
@@ -133,8 +133,21 @@ class PluginCommand(click.MultiCommand):
         if not value or ctx.resilient_parsing:
             return
 
-        click.echo('%s version %s' % (self._plugin['name'],
-                                      self._plugin['version']))
+        import qiime2.sdk
+        for entrypoint in qiime2.sdk.PluginManager.iter_entry_points():
+            plugin = entrypoint.load()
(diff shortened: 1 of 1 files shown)


ebolyen:
Sorry for losing track of this, it's on our review board now :)

PR #211, opened 2019-04-15
MAINT: New type mismatch error messages

thermokarst:


thermokarst:
Travis will fail, because we don't have new env files with the type mismatch error messages.

cc @ChrisKeefe

PR #214, opened 2019-05-20
BUG: Correctly catch out of space and does not exist errors

Oddant1:


q2cli/click/type.py
```diff
@@ -98,10 +103,28 @@ def _convert_input(self, value, param, ctx):
...
+                    temp = tempfile.tempdir
+                    self.fail(f'There was not enough space left on {temp!r} '
+                              f'to extract the artifact {value!r}. '
+                              '(Try setting $TMPDIR to a directory with '
+                              'more space, or increasing the size of '
+                              f'{temp!r})', param, ctx)
+                else:
+                    raise ControlFlowException
+            except ValueError as e:
+                if str(e) == f'{value} does not exist.':
```
ebolyen: I would probably do:
```python
if 'does not exist' in str(e):
    ...
```

to avoid the whims of the repr system

q2cli/tests/test_cli.py
```diff
@@ -282,6 +285,33 @@ def test_verbose_shows_stacktrace(self):
...
+                                            'the artifact '
+                                            f'{self.artifact1_path!r}. (Try '
+                                            'setting $TMPDIR to a directory '
+                                            'with more space, or increasing '
+                                            f'the size of {self.tempdir!r})'))
+        with self.assertRaisesRegex(click.exceptions.BadParameter,
+                                    f'{self.tempdir!r}.*'
+                                    f'{self.artifact1_path!r}.*'
+                                    f'{self.tempdir!r}'):
+            obj._convert_input(self.artifact1_path, None, None)
```
ebolyen: This last assertion isn't doing anything as the SUT (system under test) has been entirely replaced by the mock on line 300.

Oddant1: I was wondering about that, is there any better way at all to test this error? Or should this test just be cut entirely from the automated testing like we initially talked about?

ebolyen: I think [patch object](https://docs.python.org/3/library/unittest.mock.html#patch-object) does something close to what we need.

PR #217, opened 2019-06-11
NEW: Add ability to import custom click themes from config file

Oddant1:


ebolyen:
@wasade, any interest in taking this for a spin? @Oddant1, I'll start review on this shortly.

wasade:
Would love to, thanks! I will aim to do so this evening if I can swing it.

q2cli/core/config.py
```diff
@@ -0,0 +1,100 @@
...
+import configparser
+
+import click
+
+import q2cli.util
+
+
+class cli_config():
+    path = os.path.join(q2cli.util.get_app_dir(), 'cli-colors.theme')
+    valid_types = ['options', 'type', 'default_arg', 'command', 'emphasis',
```
ebolyen: What about `selectors` as a term for these? so this could be `VALID_SELECTORS`

ebolyen: and elsewhere `type` is used

q2cli/core/config.py
```diff
@@ -0,0 +1,100 @@
...
+                        val = val_user.lower()
+                        if styling == 'fg' or styling == 'bg':
+                            if val not in self.valid_colors:
+                                self.build_error(val_user, self.valid_colors,
+                                                 'valid color')
+                        else:
+                            if val not in self.valid_booleans:
+                                self.build_error(val_user, self.valid_booleans,
+                                                 'valid boolean')
+                            val = self.valid_booleans[val]
```
ebolyen: You could use [`getboolean`](https://docs.python.org/3/library/configparser.html#supported-datatypes) instead, which could replace `valid_booleans`

q2cli/core/config.py
```diff
@@ -0,0 +1,100 @@
...
+    path = os.path.join(q2cli.util.get_app_dir(), 'cli-colors.theme')
+    valid_types = ['options', 'type', 'default_arg', 'command', 'emphasis',
+                   'problem', 'errors', 'required', 'success']
+    valid_stylings = ['fg', 'bg', 'bold', 'dim', 'underline', 'blink',
+                      'reverse']
+    valid_colors = ['black', 'red', 'green', 'yellow', 'blue', 'magenta',
+                    'cyan', 'white', 'bright_black', 'bright_red',
+                    'bright_green', 'bright_yellow', 'bright_blue',
+                    'bright_magenta', 'bright_cyan', 'bright_white']
+    valid_booleans = {'true': True,
```
ebolyen: I would make these UPPERCASE since they are class-constants

q2cli/core/config.py
```diff
@@ -0,0 +1,100 @@
...
+                      'false': False,
+                      't': True,
+                      'f': False}
+    styles = {'options': {'fg': 'blue'},
```
ebolyen: This should probably be an instance variable (initialized in `__init__`) as otherwise you have to think about scope resolution between instance vs class (e.g. what does `self.styles` actually point to?), which gets strange in certain edge-cases. Also, conceptually this represents state whereas the other class-elements really aren't state, just configuration/constants.

q2cli/core/config.py
```diff
@@ -0,0 +1,100 @@
...
+import click
+
+import q2cli.util
+
+
+class cli_config():
+    path = os.path.join(q2cli.util.get_app_dir(), 'cli-colors.theme')
+    valid_types = ['options', 'type', 'default_arg', 'command', 'emphasis',
+                   'problem', 'errors', 'required', 'success']
+    valid_stylings = ['fg', 'bg', 'bold', 'dim', 'underline', 'blink',
```
ebolyen: I think you have enough of these that a set-literal may be better than a list-literal, since your operations are mostly `in`.

q2cli/core/config.py
```diff
@@ -0,0 +1,100 @@
...
+                      'reverse']
+    valid_colors = ['black', 'red', 'green', 'yellow', 'blue', 'magenta',
+                    'cyan', 'white', 'bright_black', 'bright_red',
+                    'bright_green', 'bright_yellow', 'bright_blue',
+                    'bright_magenta', 'bright_cyan', 'bright_white']
```
ebolyen: Same here re: set-literal

q2cli/tests/test_core.py
```diff
@@ -10,23 +10,27 @@
...
 from qiime2 import Artifact
 from qiime2.core.testing.type import IntSequence1
 from qiime2.core.testing.util import get_dummy_plugin
 
 import q2cli
+import q2cli.util
 import q2cli.builtin.info
 import q2cli.builtin.tools
 from q2cli.commands import RootCommand
+from q2cli.core.config import CONFIG
```
ebolyen: This can be tricky to unit-test since it's a module-level singleton. (your tests will end up side-effecting each other indirectly through the same shared object).

I would recommend constructing a new `CLIConfig` object for each test, or if you must test the module level variable, you could use [`importlib.reload`](https://docs.python.org/3/library/importlib.html#importlib.reload), but that's much more annoying to deal with.

q2cli/builtin/dev.py
```diff
@@ -29,3 +30,69 @@ def dev():
...
+@click.option('--theme', required=True,
+              type=click.Path(exists=True, file_okay=True,
+                              dir_okay=False, readable=True),
+              help='Path to file containing new theme info')
+def install_q2cli_theme(theme):
+    import os
+    import shutil
+    import q2cli.util
+    from q2cli.core.config import CONFIG
+    CONFIG.parse_file(theme)
```
ebolyen: As written, I believe an incorrect theme will become deleted from wherever I created it, which means I would need to start from scratch.

q2cli/core/config.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def __init__(self):
+        self.parse_file()
+
+    def build_error(self, current, valid_list, valid_string):
+        valids = ''
+        for valid in valid_list:
+            valids += valid + '\n'
+        raise ValueError(f'{current!r} is not a {valid_string}. The '
+                         f'{valid_string}s are:\n{valids}')
```
ebolyen: I like your error builder! :+1:

q2cli/core/config.py
```diff
@@ -0,0 +1,100 @@
...
+                if fp != self.path:
+                    raise ValueError(f'{fp!r} is not a valid filepath.')
+        except Exception as e:
+            if os.path.exists(fp):
+                os.unlink(fp)
+            raise e
+
+    def cfg_style(self, type, text, required=False):
+        reqs = {}
+        kwargs = self.styles[type]
```
wasade: Would it be possible to assert that the `type_` exists in `self.styles` or to use a `.get` followed by a test of `None` or fallback to a default style?

Oddant1: The parser doesn't generate a new style dictionary from your file, it modifies the default one, so if you didn't change any given style attribute the default value will still be in the dictionary.

q2cli/core/config.py
```diff
@@ -0,0 +1,100 @@
...
+                        self.styles[type][styling] = val
+            else:
+                # If the bad path is the default path they don't have a custom
+                # config. If the bad path isn't the default path they tried to
+                # import a nonexistent file and we want to error
+                if fp != self.path:
+                    raise ValueError(f'{fp!r} is not a valid filepath.')
+        except Exception as e:
+            if os.path.exists(fp):
+                os.unlink(fp)
```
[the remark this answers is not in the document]
Oddant1: This was a hold-over from a previous way I was doing things. At first, I was importing the file to the config path then parsing it, so if it was a bad file I wanted it removed from that location or your command line wouldn't work anymore because every time you tried to run a command it would attempt to parse the bad file and crash. Now I just make sure the file is parsable before moving it into that directory, so unlinking the file there is no longer needed. It has been changed.

q2cli/builtin/dev.py
```diff
@@ -29,3 +30,69 @@ def dev():
 def refresh_cache():
     import q2cli.core.cache
     q2cli.core.cache.CACHE.refresh()
+
+
+install_theme_help = \
+    ("Allows for customization of Qiime2's command line styling based on an "
```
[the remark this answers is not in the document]
Oddant1: I changed it to q2cli at @ebolyen's request

Oddant1:
That last commit contains an issue, I forgot that I made changes since I last tested it yesterday. I'll fix it for the next commit.

Oddant1:
Fixed

q2cli/core/config.py
```diff
@@ -0,0 +1,99 @@
...
+            self.parse_file(self.path)
+        except Exception:
+            if os.path.exists(self.path):
+                os.unlink(self.path)
+            self.styles = self._get_default_styles()
+
+    def _get_default_styles(self):
+        return {'options': {'fg': 'blue'},
+                'type': {'fg': 'green'},
+                'default_args': {'fg': 'magenta'},
```
ebolyen: I am unable to roundtrip the generate-config and install-theme as the default style uses `default_args` but the VALID_SELECTORS wants `default_arg`.

thermokarst: What about default GWAR?

q2cli/builtin/dev.py
```diff
@@ -29,3 +30,90 @@ def dev():
...
+@click.option('--theme', required=True,
+              type=click.Path(exists=True, file_okay=True,
+                              dir_okay=False, readable=True),
+              help='Path to file containing new theme info')
+def install_theme(theme):
+    import os
+    import shutil
+    import q2cli.util
+    from q2cli.core.config import CONFIG
+    CONFIG.parse_file(theme)
```
ebolyen: It would be nice to use `q2cli.util.exit_with_error` if something did go wrong (this creates a header and makes things red).

You can check out `qiime tools import` for an example of usage.

ebolyen: If it is one of your expected validation errors, we'd want to suppress the stack trace, and anything else, we probably want to print it.

q2cli/builtin/dev.py
```diff
@@ -29,3 +30,90 @@ def dev():
...
+    shutil.copy(theme, os.path.join(q2cli.util.get_app_dir(),
+                'cli-colors.theme'))
+
+
+@dev.command(name='write-default-theme',
+             short_help='Create a .ini file from the default settings at the '
```
ebolyen: Do we want the extensions for these to be `.theme` or `.ini`? We could always use language like `Create a .theme (INI formatted) file...`

q2cli/core/config.py
```diff
@@ -0,0 +1,99 @@
...
+                'emphasis': {'underline': True},
+                'problem': {'fg': 'yellow'},
+                'error': {'fg': 'red', 'bold': True},
+                'required': {'underline': True},
+                'success': {'fg': 'green'}}
+
+    def _build_error(self, current, valid_list, valid_string):
+        valids = ''
+        for valid in valid_list:
+            valids += valid + '\n'
```
ebolyen: These three lines can become just:
```
'\n'.join(valid_list)
```
That will make an empty string for an empty list, and only put newlines between the lines (saving you the double new line at the end from the final echo).

q2cli/builtin/dev.py
Oddant1: Flags can't be made required in the same way as non-flag parameters, so I did this 
```
    if not yes:
        e = ValueError("Missing required '--yes' flag.")
        header = "Missing required flag '--yes'"
        q2cli.util.exit_with_error(e, header=header, traceback=None)
```
But I don't know what to make the messages. They probably shouldn't just be the same thing like they are now

wasade: @Oddant1, would it make sense to compress this slightly? eg

```python
if not yes:
    message = "Missing required flag '--yes'"
    q2cli.util.exit_with_error(ValueError, message=message, traceback=False)
```

This would then allow `q2cli.util.exit_with_error` to construct an exception, deal with a traceback should one be desired, etc?

q2cli/core/config.py
```diff
@@ -0,0 +1,99 @@
...
+import configparser
+
+import click
+
+import q2cli.util
+
+
+class CLIConfig():
+    path = os.path.join(q2cli.util.get_app_dir(), 'cli-colors.theme')
+    VALID_SELECTORS = {'option', 'type', 'default_arg', 'command', 'emphasis',
```
wasade: Should these be `frozenset` to avoid accidental mutation?

Oddant1: That's a good idea, I'll do that

q2cli/builtin/dev.py
```diff
@@ -29,3 +30,95 @@ def dev():
...
+    from q2cli.core.config import CONFIG
+
+    CONFIG.parse_file(theme)
+    shutil.copy(theme, os.path.join(q2cli.util.get_app_dir(),
+                'cli-colors.theme'))
+
+
+@dev.command(name='write-default-theme',
+             short_help='Export the default settings.',
+             help='Create a .thenme (INI formatted) file from the default '
```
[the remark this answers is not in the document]
thermokarst:
```
if .
then
me
```

q2cli/core/config.py
```diff
@@ -0,0 +1,98 @@
...
+                      't': True,
+                      'f': False}
+
+    def __init__(self):
+        self.styles = self._get_default_styles()
+        try:
+            self.parse_file(self.path)
+        except Exception:
+            if os.path.exists(self.path):
+                os.unlink(self.path)
```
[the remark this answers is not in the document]
Oddant1: We can just reset the theme to default and raise the error through `exit_with_error` without deleting the file. The file is only being deleted from the directory we're parsing it from not wherever they imported it from, but I agree it is unnecessary to delete the file at this point.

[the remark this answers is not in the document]
Oddant1: We can also raise the error and maintain whatever their old settings were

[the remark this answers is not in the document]
ebolyen: This is in __init__, so the damage has already been done... somehow. e.g. maybe they manually edited it instead of installing it, so they skipped verification. Or perhaps they upgraded in-place (how??).

Oddant1:
Post commit https://github.com/qiime2/q2cli/pull/217/commits/d9544037711609a91a11eb9db92bee0483b408cd on top of addressing @ebolyen's review, if you attempt to import a bad file, your previous settings will be retained instead of reset to default automatically

q2cli/core/config.py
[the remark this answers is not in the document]
Oddant1: right they could theoretically dig into the file I forgot about that. Let me change some things from https://github.com/qiime2/q2cli/pull/217/commits/d9544037711609a91a11eb9db92bee0483b408cd with that in mind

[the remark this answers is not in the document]
Oddant1: @ebolyen what should the error message be if their file fails to parse when it is already in the import directory? Should we suggest they can go to `path` and fix it manually? Or just tell them to `reset-theme`. If something is wrong with that file they probably did fiddle with it manually somehow. I can't think of any other way for it to get messed up outside of some sort of data corruption.

q2cli/core/config.py
```diff
@@ -50,10 +52,23 @@ def _get_default_styles(self):
...
+    def get_editable_styles(self):
+        return {'option': {'fg': 'blue'},
+                'type': {'fg': 'green'},
+                'default_arg': {'fg': 'magenta'},
+                'command': {'fg': 'blue'},
+                'emphasis': {},
+                'problem': {'fg': 'yellow'},
+                'error': {'fg': 'red'},
+                'required': {},
+                'success': {'fg': 'green'}}
```
ebolyen: I would make all of these empty, then if someone wants zero colorization, that's possible by giving an empty file as a theme.

q2cli/builtin/dev.py
```diff
@@ -69,56 +69,71 @@ def refresh_cache():
...
     import shutil
     import q2cli.util
-    from q2cli.core.config import CONFIG
-
-    CONFIG.parse_file(theme)
+    from q2cli.core.config import ParserError, CONFIG
+
+    try:
+        CONFIG.parse_file(theme)
+    except ParserError as e:
```
ebolyen: One guy that wasn't caught with this:
```pytb
configparser.MissingSectionHeaderError: File contains no section headers.
file: 'example.theme', line: 1
'asd\n'
```

I passed a file that wasn't even plausibly a config.

q2cli/builtin/dev.py
```diff
@@ -69,56 +69,71 @@ def refresh_cache():
...
+        # If they tried to change error in a valid manner before we hit our
+        # parsing error, we don't want to use their imported error settings
+        CONFIG.styles = CONFIG.get_default_styles()
+        header = 'Something went wrong while parsing your theme: '
+        q2cli.util.exit_with_error(e, header=header, traceback=None)
```
ebolyen: This looks great! :+1:

q2cli/core/config.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def _build_error(self, current, valid_list, valid_string):
+        valids = '\n'.join(valid_list)
```
ebolyen: I think commas might look better, after seeing the fully formatted version.

q2cli/builtin/dev.py
Oddant1: I was only trying to catch the errors I raised not the errors the actual parser raised. Should I just catch all exceptions in that except?

ebolyen: Certainly not all, but I think parse errors from configparser fall under the same, expected problems which don't need a traceback category.

ebolyen: I don't know if they have a base class for all of them, but they might, which would make catching them easy.

Oddant1: Yep, `configparser.Error`, I got rid of my `ParserError` and am raising that instead, so we only need to except that error type

PR #219, opened 2019-06-18
BUG: Typo in index lookup error msg

thermokarst:


ebolyen:
Thanks @thermokarst (and @angrybee for noticing!)

PR #220, opened 2019-06-22
IMP: q2cli side of handling merging metadata with duplicate columns

Oddant1:
This is a sister PR for qiime2/qiime2#464 and will not build properly until that PR is completed and merged. This fixes a test in q2cli to reflect the fact that we will now be able to merge metadata with duplicate columns, but will still not allow for merging of duplicate files. It also starts the process of getting inspect metadata to view files with duplicated columns appropriately.

TODOS:
- `inspect-metadata` currently displays the paths and hashes regardless of if there are actually any duplicated columns, this should probably be changed so it only gives that information if there are duplicates
    - Post commit https://github.com/qiime2/q2cli/pull/220/commits/9b012c691e8cc3561fda0d9e3741a27cc0c68818 The default behavior is to only show hashes if there were duplicate column names while additionally allowing the user to specify whether they want to display hashes in all cases or not at all. Duplicate column names during the merge are tracked via the new `contains-renamed-columns` parameter in `Metadata` objects, and hash displaying is specified via the new `--show-hashes/--no-show-hashes` flag in `inspect-metadata`
- `inspect-metadata` currently does not display any sort of file/hash info if the duplicated columns come from provenance and not the files you are currently directly inspecting. Maybe this should be changed if possible?


q2cli/builtin/tools.py
```diff
@@ -248,6 +250,28 @@ def formatter(*row):
                     bold=True, nl=False)
         click.echo(metadata.column_count)
 
+        if metadata.contains_renamed_columns or show_hashes:
```
ebolyen: quick drive-by comment:

This doesn't let `no-show-hashes` avoid hashes if there are renamed columns.

Oddant1: Oh, I didn't know we wanted it to extend that far.

Oddant1: For some reason I got into my head that we wanted the default behavior to be printing the hashes whenever there were multiple files even though we had talked about not doing that before.

Oddant1:
https://github.com/qiime2/q2cli/pull/220/commits/e88144654de140ecf4b89aaac490ce748aec38a2 This just changes hash to id in the actual access of the attribute because that's what we call the parameter now, so this was needed to make the code work. All the surrounding variable names and documentation still refer to it as hash

ebolyen:
Also blocked on #277

gregcaporaso:
We've decided to close this PR, now that we have a `merge` action in q2-metadata. That action supports merging of metadata when there are either overlapping ids or overlapping columns, but not both. This was partly enabled by the addition of the `ImmutableMetadata` type, which didn't exist when work on this PR started. 

The underlying goal of this PR is still of interest (merge on the fly, without creation of a new artifact, and when there are potentially overlapping ids and overlapping columns), but it's considerably more work than adding the new action. We may circle back to reference this PR if we decide we need it in the future.

PR #224, opened 2019-07-17
BUG: Allow export to filename without relative path

### LINKED ISSUE #218: Exporting to file format fails
**Bug Description**

```bash
qiime tools export \
  --input-path table.qza \
  --output-format BIOMV100Format \
  --output-path table-v100.biom
```

Running that produces the following error:

```bash
Traceback (most recent call last):
  File "/home/matthew/.conda/envs/qiime2-2019.4/bin/qiime", line 11, in <module>
    sys.exit(qiime())
  File "/home/matthew/.conda/envs/qiime2-2019.4/lib/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/matthew/.conda/envs/qiime2-2019.4/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/matthew/.conda/envs/qiime2-2019.4/lib/python3.6/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/matthew/.conda/envs/qiime2-2019.4/lib/python3.6/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/matthew/.conda/envs/qiime2-2019.4/lib/python3.6/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/matthew/.conda/envs/qiime2-2019.4/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/matthew/.conda/envs/qiime2-2019.4/lib/python3.6/site-packages/q2cli/builtin/tools.py", line 68, in export_data
    os.makedirs(os.path.dirname(output_path), exist_ok=True)
  File "/home/matthew/.conda/envs/qiime2-2019.4/lib/python3.6/os.py", line 220, in makedirs
    mkdir(name, mode)
FileNotFoundError: [Errno 2] No such file or directory: ''
```

A way around the issue is to run with a relative path:

```bash
qiime tools export \
  --input-path table.qza \
  --output-format BIOMV100Format \
  --output-path ./table-v100.biom
```

**Computation Environment**
- OS: opensuse tumbleweed (linux)
- QIIME 2 Release 2019.4

Oddant1:
Closes #218


q2cli/tests/test_tools.py
```diff
@@ -247,6 +247,18 @@ def test_export_visualization_to_dir_success_message(self):
                   'directory %s\n' % (self.viz, output_path)
         self.assertEqual(success, result.output)
 
+    def test_export_to_filename_without_path(self):
+        current_dir = os.getcwd()
+        os.chdir(self.tempdir)
```
thermokarst: Do you need to chdir? If so, please add a comment explaining why.

q2cli/tests/test_tools.py
```diff
@@ -247,6 +247,18 @@ def test_export_visualization_to_dir_success_message(self):
                   'directory %s\n' % (self.viz, output_path)
         self.assertEqual(success, result.output)
 
+    def test_export_to_filename_without_path(self):
```
thermokarst: Can you make sure the following tests are present (add if they are not):

- export a file-format into a directory (`foo_dir/table.biom`)
- export a directory-format into the current directory (`foo_dir`)
- export a directory-format into a nested (`foo_dir/bar_dir`)

Oddant1: I'm not sure I really understand what you're asking here. Is `foo_dir` supposed to be the current working directory in all three cases? For the first case, we want to export a file-format (defined as something inheriting from `_FileFormat`?) into an actual filepath and NOT just a filename? For the second we want to export a directory-format (presumably defined in a similar manner to a file-format) into the current directory? And for the third we want to do the same as the second but export to a subdirectory of the current directory?

thermokarst: I have updated the list, for clarity.

>  Is foo_dir supposed to be the current working directory in all three cases?

No, `foo_dir` is a dir within your current working directory.

>  For the first case, we want to export a file-format (defined as something inheriting from _FileFormat?) into an actual filepath and NOT just a filename?

No, we want to export to a directory within our current working directory.

> For the second we want to export a directory-format (presumably defined in a similar manner to a file-format) into the current directory?

:+1: 

>  And for the third we want to do the same as the second but export to a subdirectory of the current directory?

:+1:

q2cli/tests/test_tools.py
thermokarst: I have an idea --- please split these new tests out into their own test class --- the chdir business can be handled in the `setUp` and `tearDown` methods, and the comment can be moved into those hooks.

q2cli/builtin/tools.py
```diff
@@ -65,6 +65,10 @@ def export_data(input_path, output_path, output_format):
                 if os.path.isfile(output_path):
                     os.remove(output_path)
                 else:
+                    # This allows the user to pass a filename as a path if they
+                    # want their output in the current working directory
+                    if os.path.dirname(output_path) == '':
+                        output_path = os.path.join('.', output_path)
```
ebolyen: Instead of adding work to be undone, what if we only ran `os.makedirs` when there _is_ a directory in the path. Otherwise, it must be a filename and the duplicate call will take care of that fine.

PR #225, opened 2019-07-23
BUG: fix completion permissions for root installed env

ebolyen:
Ran into this at STAMPS. A root conda environment (like `/opt/`) will not allow users to tab-complete as they do not have the executable bit set.

Tested manually by inspection of the flags.

cc @AstrobioMike


ebolyen:
(note, this is not urgent for STAMPS, we just used chmod after the fact which worked fine)

PR #234, opened 2019-10-28
BUG: catch missing output directories first

### LINKED ISSUE #231: Input checking should happen before the classifier runs
**Bug Description**
If you give `--o-classifier` a path to a file in a non-existent directory, it waits until after the classifier has run (sometimes several hours) before raising a FileNotFoundError.

Run `qiime feature-classifier fit-classifier-naive-bayes --o-classifier output/directory/file.qza` where `output/directory/` does not exist. 

**Expected behavior**
It should either raise a FileNotFoundError within seconds or create the directory with a warning.

**Computation Environment**
- Linux of some kind (It's a Compute Canada cluster)

ebolyen:
Fixes #231.

The code was close, but `is_writable_dir` only cares about permissions, we still need to make sure it exists.

### DIFF
--- a/q2cli/click/type.py
+++ b/q2cli/click/type.py
@@ -91,6 +91,10 @@ class QIIME2Type(click.ParamType):
                 self.fail('%r is already a directory.' % (value,), param, ctx)
 
         directory = os.path.dirname(value)
+        if directory and not os.path.exists(directory):
+            self.fail('Directory %r does not exist, cannot save %r into it.'
+                      % (directory, os.path.basename(value)), param, ctx)
+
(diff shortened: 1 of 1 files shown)


q2cli/click/type.py
```diff
@@ -91,6 +91,10 @@ def _convert_output(self, value, param, ctx):
                 self.fail('%r is already a directory.' % (value,), param, ctx)
 
         directory = os.path.dirname(value)
+        if directory and not os.path.exists(directory):
```
thermokarst: magnifique

PR #239, opened 2020-03-27
Implement CLI Usage driver

andrewsanchez:


q2cli/core/usage.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def _init_data_(self, ref, factory):
+        self._init_data_refs[ref] = factory
+        # Don't need to compute anything, so just pass along the ref
+        return ref
+
+    def _merge_metadata_(self, ref, records):
+        first_md = records[0].ref
+        remaining_records = ', '.join([r.ref for r in records[1:]])
+        t = '%s = %s.merge(%s)\n' % (ref, first_md, remaining_records)
```
thermokarst: When you get around to implementing this, its worth pointing out that metadata merging in this interface is implicit - if metadata is provided multiple times, it is merged:

```bash
--m-metadata-file foo.tsv
--m-metadata-file bar.tsv
--m-metadata-file baz.tsv
```

This will merge all three metadata files.

q2cli/core/usage.py
```diff
@@ -0,0 +1,73 @@
...
+        self._recorder.append(t)
+        return ref
+
+    def _get_metadata_column_(self, ref, record, column_name):
```
thermokarst: I suspect this will require a little but of additional machinery, behind the scenes. Like metadata merging, getting a column is inlined in the command:

```bash
--m-metadata-file foo.tsv
--m-metadata-column hello
```

So the thing to keep in mind is that in raw usage examples, the `get_metadata_column` invocation will always come before the `action` invocation - so it'll be up to you in this driver to glue those parts together, when the time comes.

q2cli/core/usage.py
```diff
@@ -0,0 +1,73 @@
...
+        t = '%s = %s.get_column(%r)\n' % (ref, record.ref, column_name)
+        self._recorder.append(t)
+        return ref
+
+    def _comment_(self, text: str):
+        self._recorder.append('# %s' % (text,))
+
+    def _action_(self, action: usage.UsageAction, input_opts: dict, output_opts: dict):
+        action_f, action_sig = action.get_action()
+        # TODO: Probably don't need this
```
thermokarst: Correct, no need to track imports, since that concept doesn't apply to this interface.

q2cli/core/usage.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def render(self):
+        return '\n'.join(self._recorder)
+
+    def get_example_data(self):
+        return {r: f() for r, f in self._init_data_refs.items()}
+
+    def _template_action(self, action_f, input_opts, outputs):
+        # The following assumes a 1-1 relationship between params and targets
+        cmd = f"qiime {action_f.plugin_id} {action_f.id}".replace("_", "-")
```
thermokarst: Might be worth looking around, I think there is some general-purpose machinery for kebabing (pinging @ebolyen).

q2cli/core/usage.py
```diff
@@ -0,0 +1,73 @@
...
+        # Might be able to use just one list
+        inputs, params = [], []
+        for k, v in input_opts.items():
+            if v in self._scope.records:
```
thermokarst: I think its possible that this approach might break down - its safer to look up the signature items on the `action_f`. Let me know if you need a hand sorting that stuff out. In general though I like how this is looking!

q2cli/tests/test_usage.py
```diff
@@ -0,0 +1,58 @@
...
+    ),
+]
+
+
+@pytest.fixture
+def dummy_plugin():
+    return get_dummy_plugin()
+
+
+@pytest.mark.parametrize("action, example, exp", params)
```
thermokarst: I'm not familiar with this feature of pytest - will it re-rerun this test multiple times with each test fixture? As long as a new instance of `CLIUsage` is constructed each time, you should be fine.

andrewsanchez: `test_render` will run once for each tuple in the list of tuples `params`.  Each tuple defines the arguments `action`, `example`, and `exp` that get passed to `test_render`.  and yes, each run will create a new instance of `CLIUsage`.  It will also provide a fresh `dummy_plugin`, but if that's unnecessary I can broaden the scope of the `dummy_plugin` pytest fixture so the same `dummy_plugin` is used throughout the module.

andrewsanchez: There is one pytest fixture in this module so far. It just sets up/tears down `dummy_plugin`.  Of course, set up/tear down happens once for each test (unless I change the scope of the fixture).  As of now, `test_render` runes twice, once for each of the two tuples in `params`.  The plan is to increase test cases by expanding `params` to cover all the test cases in `qiime2/tests/test_artifact_api.py`

q2cli/core/usage.py
```diff
@@ -0,0 +1,88 @@
...
+            for i in action_f.signature.inputs
+        ]
+
+        params = []
+        for i in action_f.signature.parameters:
+            if i in input_opts and i != "metadata":
+                spec = action_f.signature.parameters[i]
+                p = f"--p-{i.replace('_', '-')}"
+                val = input_opts[i]
+                val = f" {val}" if not isinstance(val, bool) else ""
```
thermokarst: Take a look at the machinery on `ParameterSpec` - there should be some tooling in the framework to help with this (I'll leave the specifics up as an exercise for you, @andrewsanchez).

q2cli/core/usage.py
```diff
@@ -0,0 +1,88 @@
...
+                params.append(f"{' ':>4}{p}{val}")
+        for i in self._metadata_refs:
+            p = f"--m-metadata-file"
+            val = f"{i}.tsv"
+            params.append(f"{' ':>4}{p} {val}")
+
+        # HACK: Reverse output dict for now
```
thermokarst: https://github.com/qiime2/qiime2/pull/529

q2cli/core/usage.py
```diff
@@ -0,0 +1,88 @@
...
+        rev_outputs = {v: k for k, v in outputs.items()}
+        cli_outputs = [
+            (
+                f"{' ':>4}--o-{i.replace('_', '-')} "
```
thermokarst: Please use the `to_cli_name` util.

q2cli/core/usage.py
```diff
@@ -0,0 +1,88 @@
...
+                f"{rev_outputs[i].replace('-', '_')}.qza"
```
thermokarst: Please use the `to_cli_name` util.

q2cli/core/usage.py
```diff
@@ -0,0 +1,88 @@
...
+                val = f" {val}" if not isinstance(val, bool) else ""
+                params.append(f"{' ':>4}{p}{val}")
+        for i in self._metadata_refs:
+            p = f"--m-metadata-file"
+            val = f"{i}.tsv"
+            params.append(f"{' ':>4}{p} {val}")
+
+        # HACK: Reverse output dict for now
+        rev_outputs = {v: k for k, v in outputs.items()}
+        cli_outputs = [
```
thermokarst: A personal note - list comprehensions longer than one line usually aren't worth it, IMO.

q2cli/core/usage.py
```diff
@@ -0,0 +1,88 @@
...
+        # The following assumes a 1-1 relationship between params and targets
+        cmd = f"qiime {action_f.plugin_id} {action_f.id}".replace("_", "-")
+        inputs = [
+            f"{' ':>4}--i-{i.replace('_', '-')} {input_opts[i]}.qza"
+            for i in action_f.signature.inputs
+        ]
+
+        params = []
+        for i in action_f.signature.parameters:
+            if i in input_opts and i != "metadata":
```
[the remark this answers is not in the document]
andrewsanchez: I should've guess there was a util for this!  :)

q2cli/core/usage.py
andrewsanchez: Ooh, a scavenger hunt!  😄

q2cli/core/usage.py
andrewsanchez: I wonder if there's a way to configure flake8 or something to enforce this style preference.

q2cli/core/usage.py
```diff
@@ -0,0 +1,88 @@
...
+    def get_example_data(self):
+        return {r: f() for r, f in self._init_data_refs.items()}
+
+    def _template_action(self, action_f, input_opts, outputs):
+        cmd = to_cli_name(f"qiime {action_f.plugin_id} {action_f.id}")
+        inputs, params, cli_outputs = [], [], []
+        for i in action_f.signature.inputs:
+            p = f"--i-{to_cli_name(i)}"
+            val = f"{input_opts[i]}.qza"
+            inputs.append(f"{' ':>4}{p} {val}")
```
andrewsanchez: Would you be in favor of little helper functions here, @thermokarst ?  Something like

```
inputs = template_inputs(action_f.signature.inputs)
params = template_params(action_f.signature.parameters.items)
cli_outputs = template_outputs(action_f.signature.outputs)
```

thermokarst: I like that!

Makefile
```diff
@@ -10,7 +10,7 @@ lint:
 	flake8
 
 test: all
-	QIIMETEST= nosetests
+	QIIMETEST= pytest
```
thermokarst, line 13: TODO:

Update busywork before this PR is merged:

https://github.com/qiime2/busywork/blob/162dd4ecda77c09b3ad6a6fffcfae8405c1a0158/ci/master/variables.yaml#L22

q2cli/core/usage.py
```diff
@@ -0,0 +1,114 @@
...
+class CLIUsage(usage.Usage):
+    def __init__(self):
+        super().__init__()
+        self._recorder = []
+        self._init_data_refs = dict()
+        self._metadata_refs = dict()
+        self._col_refs = dict()
+
+    def _init_data_(self, ref, factory):
+        self._init_data_refs[ref] = factory
```
thermokarst, line 27: Note to @thermokarst - this pattern has cropped up a few times - `_init_data_refs_` should be made a proper part of the API.

q2cli/core/usage.py
```diff
@@ -0,0 +1,114 @@
...
+        self._metadata_refs[record.ref] = record.result
+        self._col_refs[record.ref] = column_name
+        return ref
+
+    def _comment_(self, text: str):
+        self._recorder.append('# %s' % (text,))
+
+    def _action_(
+        self, action: usage.UsageAction, input_opts: dict, output_opts: dict
+    ):
```
thermokarst, line 47: Wow, this passes flake8?

q2cli/core/usage.py
```diff
@@ -0,0 +1,114 @@
...
+    def _assert_has_line_matching_(self, ref, label, path, expression):
+        pass
+
+    def render(self):
+        return '\n'.join(self._recorder)
+
+    def get_example_data(self):
+        return {r: f() for r, f in self._init_data_refs.items()}
+
+    def _template_action(self, action_f, input_opts, outputs):
```
thermokarst, line 62: I have a suggestion: refactor this so that signature is:

```suggestion
    def _template_action(self, action_f, inputs, parameters, metadata, outputs):
```

You can add a new utility that takes `(action_f, input_opts)` and returns `(inputs, parameters, metadata)` and call it just before you call this function. I think it might make this a little bit easier to reason about, and should help with testing, too.

q2cli/core/usage.py
```diff
@@ -0,0 +1,114 @@
...
+        cmd = to_cli_name(f"qiime {action_f.plugin_id} {action_f.id}")
+        inputs = template_inputs(action_f, input_opts)
+        params = template_parameters(action_f, input_opts)
+        params += template_metadata(self._metadata_refs, self._col_refs)
```
thermokarst, line 66: I think you might be missing some metadata cases, check the following:

https://github.com/qiime2/qiime2/blob/7061b9896708cf2f5457c76857adc64ec0d0e021/qiime2/core/testing/examples.py#L133-L142

q2cli/core/usage.py
thermokarst: Probably, but I'm not sure if its worth codifying - this is more so a general rule of thumb, IMO. Comprehensions really shine as one-liners, but because their syntax is "inverted" when compared to a traditional for loop it makes reading multi-line comprehensions tricky, because you basically have to read the comprehension backwards. Just my personal two-cents.

q2cli/core/usage.py
andrewsanchez: Lol, apparently.  I think this was `black`'s doing.  Personally, I'd be in favor of relaxing allowed line length to ~90.

q2cli/core/usage.py
andrewsanchez: Yep, those are coming soon.

q2cli/core/usage.py
andrewsanchez: 7456e36 implements your suggestion, except metadata is still consolidated into parameters.  I did it this way since metadata stuff is in `action_f.signature.parameters` but open to changing it if you have a preference.

q2cli/core/usage.py
```diff
@@ -0,0 +1,146 @@
...
+    return cli_outputs
+
+
+def examples(action):
+    all_examples = []
+    for i in action.examples:
+        use = CLIUsage()
+        action.examples[i](use)
+        example = use.render()
+        comment = f"# {i}".replace('_', ' ')
```
thermokarst, line 143: It isn't super clear where one example begins and another ends, what if you do something like:

```bash
### example 1: concatenate ints simple ########################################

# This example demonstrates basic usage.
qiime dummy-plugin concatenate-ints \
    --i-ints1 ints_a.qza \
    --i-ints2 ints_b.qza \
    --i-ints3 ints_c.qza \
    --p-int1 4 \
    --p-int2 2 \
    --o-concatenated-ints ints_d.qza

###############################################################################

### example 2: concatenate ints complex #######################################

# This example demonstrates chained usage (pt 1).
qiime dummy-plugin concatenate-ints \
    --i-ints1 ints_a.qza \
    --i-ints2 ints_b.qza \
    --i-ints3 ints_c.qza \
    --p-int1 4 \
    --p-int2 2 \
    --o-concatenated-ints ints_d.qza
# This example demonstrates chained usage (pt 2).
qiime dummy-plugin concatenate-ints \
    --i-ints1 ints_d.qza \
    --i-ints2 ints_b.qza \
    --i-ints3 ints_c.qza \
    --p-int1 41 \
    --p-int2 0 \
    --o-concatenated-ints concatenated_ints.qza

###############################################################################

### example 3: comments only ##################################################

# comment 1
# comment 2
```

q2cli/util.py
```diff
@@ -178,3 +182,22 @@ def callback(ctx, param, value):
...
+
+    def callback(ctx, param, value):
+        if not value or ctx.resilient_parsing:
+            return
+        else:
+            import q2cli.core.usage as usage
+
+        action = ctx.command._get_action()
+        examples = usage.examples(action)
+        click.secho(examples)
```
thermokarst, line 198: This should show a message when the action has no registered examples, right now it looks like this:

```bash
(q2dev) 
~/projects/qiime2/q2cli
☻  QIIMETEST= qiime dummy-plugin failing-pipeline --examples                                                                                                                                                                                                                                  cli-usage e0d1766 ✗

(q2dev) 
~/projects/qiime2/q2cli
☻      
```

q2cli/core/usage.py
```diff
@@ -0,0 +1,146 @@
...
+        self._recorder.append('# %s' % (text,))
+
+    def _action_(
+        self, action: usage.UsageAction, input_opts: dict, output_opts: dict
+    ):
+        action_f, action_sig = action.get_action()
+        inputs, params = self.extract(action_f, input_opts)
+        t = self._template_action(action_f, inputs, params, output_opts)
+        self._recorder.append(t)
+        return output_opts
```
thermokarst, line 58: Nice! This is coming along well! To elaborate a bit more on my request in the last round of review, I'm thinking something like this:

```suggestion
    def _action_(self, action, input_opts, output_opts):
        action_f, action_sig = action.get_action()
        params, mds = self._extract_from_signature(action_f, input_opts)
        t = self._template_action(action_f, input_opts, params, mds, output_opts)
        self._recorder.append(t)
        return output_opts
        
        ...
        
    def _extract_from_signature(action_f, input_opts):
        ...
        return (extracted_params, extracted_mds)
        
    def _template_action(action_f, inputs, params, mds, outputs):
        inputs_t = self._template_inputs(action_f, inputs)
        params_t = self._template_parameters(action_f, params)
        mds_t = self._template_metadata(action_f, mds)
        outputs_t = self._template_outputs(action_f, outputs)
        
        cmd = to_cli_name(f"qiime {action_f.plugin_id} {action_f.id}")
        t = " \\\n".join([cmd] + inputs_t + params_t + mds_t + outputs_t)
        return t
```

I think with a structure like this in place you might be able to reduce the need for the state attributes tacked onto the class (but we can talk about that more in the next round).

So, rephrasing the thought above, we might see an API like this:

```python
class CLIUsage(usage.Usage):
    # from parent class
    def __init__(self): pass
    def _init_data_(self, ref, factory): pass
    def _merge_metadata_(self, ref, records): pass
    def _get_metadata_column_(self, ref, record, column_name): pass
    def _comment_(self, text: str): pass
    def _action_(self, action, input_opts, output_opts): pass
    def _assert_has_line_matching_(self, ref, label, path, expression): pass
    # for this child class
    ## private api
    def _extract_from_signature(self, action_f, input_opts): pass
    def _template_action(self, action_f, inputs, params, mds, outputs): pass
    def _template_inputs(self, action_f, inputs): pass
    def _template_parameters(self, action_f, params): pass
    def _template_metadata(self, action_f, mds): pass
    def _template_outputs(self, action_f, outputs): pass
    ## public api
    def render(self): pass
    def get_example_data(self): pass
```

q2cli/core/cache.py
```diff
@@ -265,13 +265,19 @@ def _get_plugin_state(self, plugin):
 
     def _get_action_state(self, action):
         import itertools
+        import pickle
 
+        examples = {
+            k: pickle.dumps(v, 0).decode()
+            for k, v in action.examples.items()
+        }
```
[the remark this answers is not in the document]
thermokarst: Have a peek at:

https://github.com/qiime2/q2cli/compare/master...ebolyen:usage

His solution there was to push the rendering up into the caching layer, which I think will work well for your PR, too.

thermokarst:
I'm gonna keep going on my review of the main usage driver, but you should be free to continue working on the caching business.

q2cli/core/usage.py
```diff
@@ -0,0 +1,156 @@
...
+        params, mds = [], []
+        for i, spec in action_sig.parameters.items():
+            q_type = spec.qiime_type
+            if is_metadata_type(q_type):
+                mds.append((i, spec))
+            else:
+                params.append((i, spec))
+        return params, mds
+
+    def _template_action(self, action, input_opts, outputs):
```
thermokarst, line 75: A general idea for the templating methods: take a look at the [`textwrap`](https://docs.python.org/3/library/textwrap.html) module in the stdlib, it would help to just have one pass on formatting indentation at the end of assembling the commands.

q2cli/core/usage.py
```diff
@@ -0,0 +1,156 @@
...
+        action_f, action_sig = action.get_action()
+        cmd = to_cli_name(f"qiime {action_f.plugin_id} {action_f.id}")
+        params, mds = self._extract_from_signature(action_sig)
+        inputs_t = self._template_inputs(action_sig, input_opts)
+        params_t = self._template_parameters(params, input_opts)
+        mds_t = self._template_metadata(mds, input_opts)
+        outputs_t = self._template_outputs(action_sig, outputs)
+        t = " \\\n".join([cmd] + inputs_t + params_t + mds_t + outputs_t)
```
thermokarst, line 83: Restating what I mentioned above in this review - this line could be refactored as another helper method - it takes the raw command elements and returns an indented, escaped command.

q2cli/core/usage.py
```diff
@@ -0,0 +1,156 @@
...
+                    for k, v in data.items():
+                        mds_t.append(f"{' ':>4}{file_param} {k}.tsv")
+            elif is_metadata_column_type(qtype):
+                col, md = input_opts[i]
+                mds_t.append(f"{' ':>4}{file_param} {md}.tsv")
+                mds_t.append(f"{' ':>4}{col_param} '{col}'")
+        return mds_t
+
+
+def examples(action):
```
thermokarst, line 147: This should be refactored to use the cached values, now that you have that available to you.

q2cli/core/usage.py
```diff
@@ -0,0 +1,156 @@
...
+
+class CLIUsage(usage.Usage):
+    def __init__(self):
+        super().__init__()
+        self._recorder = []
+        self._init_data_refs = dict()
+
+    def _init_data_(self, ref, factory):
+        self._init_data_refs[ref] = factory
+        data = factory()
```
thermokarst, line 31: This could potentially be expensive (especially for poorly-behaved usage examples). I think it would be fine to not do these type checks, and instead below in `_template_metadata` you can just invert the logic there - lean on the `signature` (the signature can tell you which of your input opts _should_ be metadata). 

Ultimately the typechecking will be handled by the framework, when a usage example is exercised with the `ExecutionUsage` driver.

andrewsanchez: I was trying to ensure metadata column names are accessible.  When a metadata column is passed in as a factory, e.g.,

```python
def identity_with_metadata_column_from_factory(use):
    ints = use.init_data('ints', ints1_factory)
    mdc = use.init_data('mdc', mdc1_factory)


    use.action(
        UsageAction(plugin_id='dummy_plugin',
                    action_id='identity_with_metadata_column'),
        UsageInputs(ints=ints, metadata=mdc),
        UsageOutputNames(out='out'),
    )
```
I think the only way to get the column name is to call the factory.

Perhaps the solution is to only call a factory when it's a metadata column factory.

thermokarst: Gotcha - the `ref` in the `_init_data_` signature is the MDC's name, so you can just use the ref as-is. Another reason for us to get docstrings into the usage API!

andrewsanchez: So in the case above, your saying the MDC's name for `mdc1_factory` is `'mdc'`?

I thought that the MDC's name for the `mdc1_factory` was `'a'`.  Is that not what we're after?  
 
```python
def mdc1_factory():
    return CategoricalMetadataColumn(pd.Series(['1', '2', '3'],
                                     name='a',
                                     index=pd.Index(['0', '1', '2'],
                                                    name='id')))
```

I think getting the actual name of the column is only possible after calling the factory.  Am I wrong about that?  I feel like I might be misunderstanding something here.

thermokarst: No worries, I misspoke, I was thinking about `get_metadata_column`, but was talking about `_init_data_` (`get_metadata_column` includes a ref to the column name for just this case). I think this might be a good reason to not support the case you pointed out above (`identity_with_metadata_column_from_factory`). I recall discussing this with @ebolyen, but I don't recall any of the specifics, but anyway, I think that maybe this is an indication that maybe the API needs to be tweaked a bit.

thermokarst: BTW, @ebolyen, I think when we chatted about this before, I referred to it as "the least common denominator" issue - since we _must_ have a metadata file for interfaces like q2cli, maybe that means that Usage API needs to follow a similar guideline of working against the "least common denominator" (which is `Metadata` only, not `Metadata` and `MetadataColumn`).

@andrewsanchez - what do you think of this?

andrewsanchez: What if we change the signature of `init_data` to something like this:

```diff
modified   qiime2/core/testing/examples.py
@@ -193,7 +193,7 @@ def identity_with_metadata_column_get_mdc(use):
 
 def identity_with_metadata_column_from_factory(use):
     ints = use.init_data('ints', ints1_factory)
-    mdc = use.init_data('mdc', mdc1_factory)
+    mdc = use.init_data('mdc', mdc1_factory, name='a')
 
     use.action(
         UsageAction(plugin_id='dummy_plugin',
modified   qiime2/sdk/usage.py
@@ -217,8 +217,8 @@ class Usage(metaclass=abc.ABCMeta):
     def __init__(self):
         self._scope = Scope()
 
-    def init_data(self, ref, factory):
-        value = self._init_data_(ref, factory)
+    def init_data(self, ref, factory, name=None):
+        value = self._init_data_(ref, factory, name)
```

Then CLIUsage can do something like this:
```diff
modified   q2cli/core/usage.py
@@ -28,8 +28,10 @@ class CLIUsage(usage.Usage):
         self._recorder = []
         self._init_data_refs = dict()
 
-    def _init_data_(self, ref, factory):
+    def _init_data_(self, ref, factory, name=None):
         self._init_data_refs[ref] = factory
+        if name:
+            return name, ref
         return ref
@@ -144,20 +146,15 @@ class CLIUsage(usage.Usage):
                     for k, v in data.items():
                         mds_t.append(f"{file_param} {k}.tsv")
             elif is_metadata_column_type(qtype):
-                try:
-                    col, md = input_opts[i]
-                except ValueError:
-                    # The metadata column was passed in as a factory. The
-                    # factory needs to be called so we can get the column name
-                    md = input_opts[i]
-                    data = self._init_data_refs[md]()
-                    col = data.name
+                col, md = input_opts[i]
                 mds_t.append(f"{file_param} {md}.tsv")
                 mds_t.append(f"{col_param} '{col}'")
         return mds_t

```

thermokarst: That's an interesting idea - do you think that would be confusing for the other types of data, like Artifacts? I'm kinda thinking that maybe we should refactor `init_data` into two or three new methods: `init_artifact`, `init_metadata`, and `init_metadata_column`. The factories would return data to be fed into the `Artifact`, `Metadata`, and `MetadataColumn` constructors, rather than instances of the aforementioned data types. Then, the scope can actually keep track of the types of the data, which then would let the interface drivers do something specific with that info. For example, with this `identity_with_metadata_column_from_factory` case, the `CLIUsage` driver could actually just "materialize" a new `Metadata` tsv file with one column, the column used and referenced in the `init_metadata_column` dataset.

andrewsanchez: > That's an interesting idea - do you think that would be confusing for the other types of data, like Artifacts? 

With sufficient documentation, I don't think so.   I think my proposal is a little hacky, but perhaps not terribly so?  Your proposal definitely seems better but maybe more of a "nice to have," worth implementing to improve the API later?  Since my proposal is such a small, non-breaking* change, perhaps we can use it as an interim solution and open an issue for an API improvement later.

It would be non-breaking with the addition of something like:

```diff
+import inspect
 import re
 import types
 
@@ -217,11 +218,14 @@ class Usage(metaclass=abc.ABCMeta):
     def __init__(self):
         self._scope = Scope()
 
-    def init_data(self, ref, factory):
-        value = self._init_data_(ref, factory)
+    def init_data(self, ref, factory, name=None):
+        if "name" in inspect.getargspec(self._init_data_).args:
+            value = self._init_data_(ref, factory, name)
+        else:
+            value = self._init_data_(ref, factory)
         return self._push_record(ref, value)
 
-    def _init_data_(self, ref, factory):
+    def _init_data_(self, ref, factory, name=None):
         raise NotImplementedError
 
     def merge_metadata(self, ref, *records):
```

q2cli/util.py
```diff
@@ -178,3 +182,26 @@ def callback(ctx, param, value):
...
+
+        action = ctx.command._get_action()
+        plugin_id = to_cli_name(action.plugin_id)
+        cached_plugin = q2cli.core.cache.CACHE.plugins[plugin_id]
+        cached_action = cached_plugin['actions'][action.id]
+        msg = cached_action.get("examples")
+        if not msg:
+            msg = "No examples have been registered for this action yet."
+        click.secho(msg)
+        ctx.exit()
```
andrewsanchez, line 203: @thermokarst is this sufficient for leveraging rendered usage examples that have been cached?

q2cli/core/usage.py
andrewsanchez: If I'm not mistaken, I think the the click option is the only place that needs to know about the cache.  Usage examples will only be rendered via `q2cli.core.usage.examples` when the cache is refreshed as per the pre-existing logic.  See my inline comment in `q2cli/util.py`

q2cli/core/usage.py
thermokarst: > With sufficient documentation, I don't think so

Interesting! I'm curious what you have in mind. The case that I'm wondering about can be summarized like this:

```python
def ints1_factory():
    return Artifact.import_data(IntSequence1, [0, 1, 2])

def concatenate_ints_simple(use):
    ints_a = use.init_data('ints_a', ints1_factory, name='spleen')
   ...
```

What does the name mean in this case? How does it differ from the `ints_a` name (ref) that is being specified?

> but maybe more of a "nice to have," worth implementing to improve the API later?

We aren't in a huge rush, I don't mind ironing out the API a bit more. When we merged Artifact API into 2020.2 we recognized that the API wasn't quite complete, and we would deal with issues as they cropped up.

q2cli/util.py
thermokarst: I think so! I'll take a closer look soon, but this looks reasonable to me. Thanks!

thermokarst:
@andrewsanchez - I just cut a draft PR ([link](https://github.com/qiime2/qiime2/pull/536)), I think it will help provide you with some new options for dealing with metadata. The new PR on the framework proposes to add a `source` property to to the usage scope records. You can use those values to identify where and how a record was created. So, for your metadata processing, you can check to see:
- if the source is `init_data`, and it is being passed directly into a MDC param, then it must be a MDC
- if the source is `init_data`, and it is being passed directly into a MD param, then it must be a MD
- if the source is `merge_metadata`, then it must be metadata
- if the source is  `get_metadata_column`, then it must be metadata

We don't need to be 100% type-checked here - the assumption is that usage examples first pass the `ExecutionUsage` "check", which _does_ ensure type-checking.

Let me know if you have any questions!

andrewsanchez:
> * if the source is `init_data`, and it is being passed directly into a MDC param, then it must be a MDC

When you say "passed into a [MD|MDC] param," are you referring to the return value of `init_data` being passed into `UsageInputs`?  Both metadata and metadata columns are passed into `UsageInputs` as kwargs and the examples pass all metadata to `UsageInputs` under the kwarg `metadata`.  So I'm unsure how I would know that it's a MDC when the source is `init_data`.

It seems to me that I the `source` property addresses all but the identity_with_metadata_column_from_factory` example.  If I get a MDC from a factory, it just looks like MD.  I can't tell that it's a MDC.  Even if I _did_ know that it was a MDC, I would then need to call the factory to get the correct column name.

So I have one failing test case here, with tests run against qiime2/qiime2#536

thermokarst:
> are you referring to the return value of init_data being passed into UsageInputs?

No, not quite - sorry, my message above is a bit confusing. I was referring to the the action signature - if the param that was _receiving_ the `ScopeRecord` with an `init_data` source had a `qiime_type` of MD|MDC, then it must be a MD|MDC. Yeesh, re-reading that, that too is a mouthful. Okay, third time should get it, right:

```python
           if qiime2.sdk.util.is_metadata(spec.qiime_type):
                if source not in {'init_data', 'merge_metadata'}:
                    raise TypeError('this is looking like mdc to me')
            elif qiime2.sdk.util.is_metadata_column(spec.qiime_type):
                #do something here?
            else:
                raise TypeError('how did i wind up in this place?')
```

Maybe that untested code conveys the message?

---

> If I get a MDC from a factory, it just looks like MD. I can't tell that it's a MDC.

This is where you interrogate the signature - figure out what the param's type is, first. Then you can start making assumptions about the scope record.

> Even if I did know that it was a MDC, I would then need to call the factory to get the correct column name.

Let's stick with convention here - the `ref` name that is assigned to the scope record should match the column name.

andrewsanchez:
To think of all the times I have failed to use the tools available to me, right before my eyes  😭

andrewsanchez:
> Let's stick with convention here - the `ref` name that is assigned to the scope record should match the column name.

Wouldn't this cause problems later if there is a mismatch between a rendered CLI Usage example and the _actual_ column name?

thermokarst:
> To think of all the times

lol

> Wouldn't this cause problems later if there is a mismatch between a rendered CLI Usage example and the actual column name? Or is the expectation that the author of an example will make sure the names and refs match up?

@andrewsanchez and I chatted about this IRL and I told him "yes" and "yes" - I think right now my expectation is that Usage examples are canonically validated by running them through ExecutionUsage.

Another idea crossed my mind - we could also invert power dynamic! Specifically, your usage drive can actually make a MDC have whatever name it darn well wants - the driver can just update the MDC name when it saves the example data to disk (the think @ebolyen asked you about in our lab mtg the other day). I'm gonna let that percolate for a bit, but it shouldn't change any of today's IRL discussion conclusions, just something to chew on for the future.

q2cli/core/usage.py
```diff
@@ -0,0 +1,161 @@
...
+    def _template_metadata(self, mds, input_opts):
+        mds_t = []
+        file_param = "--m-metadata-file"
+        col_param = "--m-metadata-column"
+        scope_records = self._scope.records
+        for i, spec in mds:
+            name = input_opts[i]
+            if not isinstance(name, str):
+                name, result = name
+            record = scope_records[name]
```
thermokarst, line 134:
```suggestion
        for i, spec in mds:
            name = input_opts[i]
            if not isinstance(name, str):
                name, result = name
            record = self._get_record(name)
```

andrewsanchez:
@thermokarst, regarding your requests above for handling collections in the `_template_inputs` and `_template_parameters` methods, I think [this example](https://github.com/qiime2/qiime2/blob/81f9bc1e735a1d9c1c852d144b8302ab4c24a3cc/qiime2/core/testing/examples.py#L186-L195) covers both of those cases:  

```python
def variadic_input_simple(use):
    ints = use.init_data('int', int_sequence_list_factory)
    int_set = use.init_data('int_set', single_int_set_factory)


    use.action(
        UsageAction(plugin_id='dummy_plugin',
                    action_id='variadic_input_method'),
        UsageInputs(ints=ints, int_set=int_set, nums={7, 8, 9}),
        UsageOutputNames(output='out'),
    )
```

However, while I can check `is_collection_type`, isn't the information needed to render the CLI example available only in the factory?

The factories in question are:

```python
def int_sequence_list_factory():
    int1 = ints1_factory()
    int2 = ints2_factory()
    return [int1, int2]

def single_int_set_factory():
    single_int1 = single_int1_factory()
    single_int2 = single_int2_factory()
    return {single_int1, single_int2}
```

Please correct me if I am wrong, but as far as I can tell, there's no way of knowing how many additional `--i-` or `--p-` flags I would need without knowing the length of the return value of these factories.

thermokarst:
> there's no way of knowing how many additional --i- or --p- flags I would need without knowing the length of the return value of these factories.

Good catch! Yeah, Usage API should keep track of that. Here is the first pass of a diff on the framework to implement this kind of tracking:

```patch
diff --git a/qiime2/core/testing/examples.py b/qiime2/core/testing/examples.py
index 6c708c6..de338e7 100644
--- a/qiime2/core/testing/examples.py
+++ b/qiime2/core/testing/examples.py
@@ -43,12 +43,6 @@ def md2_factory():
                                                 name='id')))
 
 
-def int_sequence_list_factory():
-    int1 = ints1_factory()
-    int2 = ints2_factory()
-    return [int1, int2]
-
-
 def single_int1_factory():
     return Artifact.import_data(SingleInt, 10)
 
@@ -184,7 +178,7 @@ def identity_with_metadata_column_get_mdc(use):
 
 
 def variadic_input_simple(use):
-    ints = use.init_data('int', int_sequence_list_factory)
+    ints = use.init_data_collection('int', list, ints1_factory, ints2_factory)
     int_set = use.init_data('int_set', single_int_set_factory)
 
     use.action(
diff --git a/qiime2/sdk/usage.py b/qiime2/sdk/usage.py
index af4e3be..81a655c 100644
--- a/qiime2/sdk/usage.py
+++ b/qiime2/sdk/usage.py
@@ -231,6 +231,16 @@ class Usage(metaclass=abc.ABCMeta):
     def _init_data_(self, ref, factory):
         raise NotImplementedError
 
+    def init_data_collection(self, ref, container_type, *records):
+        if len(records) < 2:
+            raise ValueError('Must provide two or more Artifact inputs.')
+
+        value = self._init_data_collection_(ref, container_type, records)
+        return self._push_record(ref, value, 'init_data_collection')
+
+    def _init_data_collection_(self, ref, container_type, records):
+        raise NotImplementedError
+
     def init_metadata(self, ref, factory):
         value = self._init_metadata_(ref, factory)
         return self._push_record(ref, value, 'init_metadata')
```

Can you cut a PR on the framework that finishes implementing those changes? Let me know if you need a hand.

andrewsanchez:
@thermokarst, can you check the rendered example, please?  🙏 

# Usage Example
```python
def variadic_input_simple(use):
    ints = use.init_data_collection('ints', list, int1=ints1_factory,
                                    int2=ints2_factory)
    int_set = use.init_data_collection('int_set', set,
                                       single_int1=single_int1_factory,
                                       single_int2=single_int2_factory)

    use.action(
        UsageAction(plugin_id='dummy_plugin',
                    action_id='variadic_input_method'),
        UsageInputs(ints=ints, int_set=int_set, nums=[7, 8, 9]),
        UsageOutputNames(output='out'),
    )
```

# CLIUsage rendering
```
qiime dummy-plugin variadic-input-method \
    --i-ints int1.qza \
    --i-ints int2.qza \
    --i-int-set single_int1.qza \
    --i-int-set single_int2.qza \
    --p-nums 7 \
    --p-nums 8 \
    --p-nums 9 \
    --o-output out.qza
```

q2cli/tests/test_usage.py
```diff
@@ -0,0 +1,227 @@
...
+            "    --o-output output.qza",
+            "qiime dummy-plugin optional-artifacts-method \\",
+            "    --i-ints ints.qza \\",
+            "    --p-num1 1 \\",
+            "    --p-num2 2 \\",
+            "    --o-output output.qza",
+            "qiime dummy-plugin optional-artifacts-method \\",
+            "    --i-ints ints.qza \\",
+            "    --p-num1 1 \\",
+            "    --p-num2 None \\",
```
andrewsanchez, line 166: @thermokarst do we actually want to pass `None` here?

thermokarst:
RE the proposed API here:

```python
    ints = use.init_data_collection('ints', list, int1=ints1_factory,
                                    int2=ints2_factory)
```

This looks pretty nice! I have one concern, though. How will this work for data that is coming from an upstream action? I think that is another reason why working with the scope records is the right way to go, rather than with the factories.

q2cli/tests/test_usage.py
thermokarst: Not if it wasn't explicitly set in the usage example.

andrewsanchez:
> RE the proposed API here:
> 
> ```python
>     ints = use.init_data_collection('ints', list, int1=ints1_factory,
>                                     int2=ints2_factory)
> ```
> 
> This looks pretty nice! I have one concern, though. How will this work for data that is coming from an upstream action? I think that is another reason why working with the scope records is the right way to go, rather than with the factories.

I'm not sure what you mean.  There are no scope records at this point.  At what point are you referring to working with the scope records?  The return value of `init_data_collection` dictates the contents of the scope record for the collection.  There are no scope records for the collection members.  Changing the signature to require kwargs gives me access to the refs for each member of the collection which I can then dump into to the `result` attribute of the record for the collection.  So I'm just letting `UsageInputs.build_opts` process the records and then grab the info when it comes through in the templating methods.

thermokarst:
>  At what point are you referring to working with the scope records?

The input values for `init_data_collection` should be scope records, not factories. Otherwise, you're in the exact same boat as before - not enough info to get the job done.

Here is an example where use factories breaks down:

```python
def ints1_factory():
    return Artifact.import_data(IntSequence1, [0, 1, 2])

def md1_factory():
    return Metadata(pd.DataFrame({'a': ['1', '2', '3']},
                                 index=pd.Index(['0', '1', '2'],
                                                name='id')))

def failing_example(use):
    ints = use.init_data('ints', ints1_factory)
    md = use.init_metadata('md', md1_factory)

    use.action(
        UsageAction(plugin_id='dummy_plugin',
                    action_id='identity_with_metadata'),
        UsageInputs(ints=ints, metadata=md),
        UsageOutputNames(out='out'),
    )

    out_ints = use.get_result('out')

    # What do I do with `out_ints` here? It is a scope record, not a factory.
    ints = use.init_data_collection('ints', list, int1=ints1_factory,
                                    int2=ints2_factory)

    use.action(
        UsageAction(plugin_id='dummy_plugin',
                    action_id='variadic_input_method'),
        UsageInputs(ints=ints, int_set=int_set, nums={7, 8, 9}),
        UsageOutputNames(output='out'),
    )
```

> How will this work for data that is coming from an upstream action?

This example is what I was talking about in that quote.

thermokarst:
PS - sorry for the confusion. I was the one who originally proposed using factories (instead of scope records; https://github.com/qiime2/q2cli/pull/239#issuecomment-640858209), but realized that was a mistake in https://github.com/qiime2/qiime2/pull/539#discussion_r437562349

andrewsanchez:
> The input values for `init_data_collection` should be scope records, not factories. Otherwise, you're in the exact same boat as before - not enough info to get the job done.
> 
> Here is an example where using factories breaks down:

Thank you for the example!   However, I am still confused because I was under the impression that  `init_data_collection`  **_has to handle factories_**.  Is that not the case?  I thought your original suggestion for passing into records was a typo.

What about extending this method to handle scope records _and_ factories?

thermokarst:
> Thank you for the example!

❤️ No, _thank you_

> has to handle factories. Is that not the case?

No, I don't think so - scope records will contain a ref, and if the driver implementation needs it, the factory - so its the best of both worlds. For a similar example, check out `merge_metadata`:

https://github.com/qiime2/qiime2/blob/57827c83b39823f497c5f05454cd605edbb77d95/qiime2/sdk/usage.py#L241

> I thought your original suggestion for passing into records was a typo.

You're right - there was a typo: I was passing in factories, but referred to them as "records," because I was copying the sig for `merge_metadata`. Ooops! To top it off, the typo was actually nudging me in the right direction - scope records all the way!

> What about extending this method to handle scope records and factories? 

That sounds technically possible, but let me ask you - how does that impact the plugin-dev experience? Also, what happens if someone does something silly like this:

```python
out_ints = use.get_result('out') # declared ref is "out", which means the cli name will be out.qza
ints = use.init_data_collection('ints', list, spleen=out_ints, int2=ints2_factory) # uh oh
```

What do we do there? Do we name it `spleen.qza`? Or `out.qza`? If we go with `spleen.qza`, we have to retroactively rename _every spleen reference_. If we go with `out.qza`, well, then why did we ask the user to give us a new name in the first place?

Ack!

thermokarst:
Another thought popped into my mind - the name I proposed for this function (`init_data_collection`) is probably a bit confusing now, too. Maybe we should take a page out of `merge_metadata`'s book and call this something like `combine_data` or `collect_data` or ... something.

Makefile
thermokarst: @andrewsanchez - please cut a PR on qiime2/busywork that addresses the change requested above - thanks!

andrewsanchez: Looks like this is outdated.  Is there still something that needs to be done?  Should I remove  Here are all references to `nosetests`:

image

Should I remove nose from the q2cli runners?

andrewsanchez: https://github.com/qiime2/busywork/pull/102

thermokarst:
Looks great @andrewsanchez! You have done a spectacular job here, congrats!

I have pushed up a handful of commits (meant to be reviewed separately, but in order - they build on each other):

- b6cf4b0433de1df120f5037786a6188c3f338170: Unifies the data-flow, basically taking all of the one-off data prep steps, and unifying into a few new methods (`_destructure_signature` and `_destructure_opts`) - these two handle the heavy lifting, so to speak. Also, for the sake of simplicity, I changed up the rendering of Boolean parameters, which allows us to unify handling of param templates.
- ed9bcedddee3c5640581e8e85ca6e54e688dc7ab: Begins the process of refactoring the caching vs templating of the examples. Basically, I think we need to store something more like an AST in the cache, rather than the fully rendered examples. In the future we will want to be able to parameterize the templating (e.g. "styles", like "save all outputs into an output dir" as a style, or "write command on a single line" style). By storing the ASTish data in the cache (rather than fully rendered commands), we get the best of both worlds - fast caching, and flexible rendering.
- 69fad2427315aac2c22a5b21e82ad20916adf02d: Finishes what was started in ed9bcedddee3c5640581e8e85ca6e54e688dc7ab: splits the `CLIUsage` into two classes, one that writes to the cache, and one that consumes from the cache, basically.
- b6cb3c3f91214fa67480a5f336f74d8e2a3896f9: linting. Standardizing on single quotes (there was a mix of single and double in here).

Please review when you get a chance and let me know what you think! There are still some TODOs hanging around, including dealing with parameter collections, which means we probably need more tests. I can take point on all of that, if you want, please just let me know.

ci/recipe/meta.yaml
```diff
@@ -18,6 +18,7 @@ requirements:
   host:
     - python  {{ python }}
     - setuptools
+    - pytest
```
thermokarst, line 21: Please add a comment explaining why this dep is necessary (just so that future readers don't remove this because they think its a mistake and remove it - I think this is the only case where we specify pytest as a direct dependency for a plugin).

q2cli/core/usage.py
```diff
@@ -0,0 +1,231 @@
...
+        ]
+
+        base_cmd = to_cli_name(f'qiime {plugin_id} {action_id}')
+        action_t = self._format_templates(base_cmd, templates)
+        return action_t
+
+    def _format_templates(self, command, templates):
+        wrapper = textwrap.TextWrapper(initial_indent=' ' * 4)
+        templates = [command] + [wrapper.fill(t) for t in templates]
+        # TODO: double-check that string escaping is working
```
thermokarst, line 162: I'm not 100% sure that all of our string escaping et al is set up quite right - I need to think more on this.

q2cli/core/usage.py
```diff
@@ -0,0 +1,231 @@
...
+            refs = ref if isinstance(ref, list) else [ref]
+            for ref in refs:
+                opt_name = to_cli_name(opt_name)
+                inputs.append(f'--i-{opt_name} {ref}.qza')
+        return inputs
+
+    def _template_parameters(self, param_opts):
+        params = []
+        for opt_name, (val, _) in param_opts.items():
+            # TODO: circle back on sets
```
thermokarst, line 177: Native Python sets aren't JSON compatible, so prior to serializing to JSON for the CLI Cache, we convert any collections into `list` (https://github.com/qiime2/q2cli/pull/239/files#diff-0e5556f314d00b6f2414122af6ed2260R103-R104). The templater here will need to unpack that info. Strategies include looking at the `qiime_type` (its the `_` above), or, serializing the original collection type (`_destructure_opts`), or, somehow looking up the record (you would need to do that in `CLIUsage` though, the records will no longer exist here in `CLIRenderer`).

q2cli/tests/test_usage.py
```diff
@@ -0,0 +1,257 @@
...
+    rendered = '\n'.join(list(cache_renderer.render()))
+    assert rendered == '\n'.join(exp)
+
+
+class CliTests(unittest.TestCase):
+    def setUp(self):
+        get_dummy_plugin()
+        self.runner = CliRunner()
+
+    def test_examples(self):
```
thermokarst, line 219: I just realized I put this test in the wrong place (I think) - if you agree, can you move over to `test_cli.py`? Once moved, this whole class (`CliTests`) can be removed.

q2cli/tests/test_usage.py
```diff
@@ -0,0 +1,257 @@
...
+from qiime2.core.testing.util import get_dummy_plugin
+
+
+@pytest.fixture
+def dummy_plugin(monkeypatch):
+    monkeypatch.setenv('QIIMETEST', '')
+    return get_dummy_plugin()
+
+
+params = [
```
thermokarst, line 25: Regarding test cases - these all look good - can you add a test for _every other usage example_ in the dummy plugin?

https://github.com/qiime2/qiime2/blob/master/qiime2/core/testing/examples.py#L54-L223

These might be good to start with, that way you can TDD your way into any minor tweaks that might need to happen in the templater (for example, the `set` business I mentioned above).

q2cli/util.py
```diff
@@ -178,3 +182,23 @@ def callback(ctx, param, value):
     return click.Option(['--citations'], is_flag=True, expose_value=False,
                         is_eager=True, callback=callback,
                         help='Show citations and exit.')
+
+
+def usage_example_option(action):
```
thermokarst, line 187: It would be really cool to have an `--example-data ./foo_dir` command, that would render out the usage data. I'm not so sure what the best way is to do this, though, since we won't have references to the data anymore. Options include:

- make a one-off `ExecutionUsage`-style usage driver, that gets invoked directly when rendering the data - this would allow us to bypass the cache
- save the example data in the cache when we initialize the cache. that sounds like it will backfire, but I am open to it.

Please leave this until the end - this is more of a "nice-to-have" than a "must-have" for getting this feature merged.

q2cli/core/usage.py
```diff
@@ -0,0 +1,231 @@
...
+from qiime2.sdk.util import (
+    is_metadata_type,
+    is_visualization_type,
+)
+
+from q2cli.util import to_cli_name
+
+
+def is_iterable(val):
+    return isinstance(val, collections.abc.Iterable)
```
andrewsanchez, line 22: Love this!  Didn't research this too deeply, but I never found a satisfying answer for reliably performing this check.

q2cli/tests/test_usage.py
andrewsanchez: The only missing usage example I see is `comments_only`.

thermokarst: Perfect, thanks for checking - can you add the `comments_only` case? Thanks!

thermokarst: Resolved in d5fc3e7

q2cli/tests/test_cli.py
```diff
@@ -340,6 +340,76 @@ def test_run_deprecated_gets_warning_msg(self):
...
+        )
+
+        exp = (
+            '# ### example: identity with metadata column get mdc ###',
+            'qiime dummy-plugin identity-with-metadata-column \\',
+            '    --i-ints ints.qza \\',
+            '    --m-metadata-file md.tsv \\',
+            '    --m-metadata-column \'a\' \\',
+            '    --o-out out.qza\n'
+        )
```
andrewsanchez, line 396: This example was not rendering properly after reading from the cache.  We were making an implicit assumption about columns based on an `isinstance(ref, tuple)` check, which didn't work since tuples are not serializable as JSON.

Please see my proposed solution in f6ca1440b53953e4b58e7894dc9483b18a196752

q2cli/util.py
thermokarst: Let's save this for 2020.8, I looked at what we would need to do to implement this, and I am not comfortable adding that much functionality this close to the release.

PR #252, opened 2021-02-09
refactor: usage renderer logic is now fully priv

thermokarst:
this work came up in a discussion with @andrewsanchez - having a `render` method on all of the usage drivers would help simplify his life when building sphinx-ext-qiime2. I plan to follow up to make that `render` method part of the base `Usage` class API.

Also, while working on this I had some ideas for refactoring the CLI-specific usage cache, more on that later.

### DIFF
--- a/q2cli/core/usage.py
+++ b/q2cli/core/usage.py
@@ -22,12 +22,24 @@ class CLIUsage(usage.Usage):
         super().__init__()
         self._cache_recorder = []
 
+    @property
     def cache(self):
         return self._cache_recorder
+
+    @classmethod
+    def from_cache(cls, cache):
+        instance = cls()
+        instance._cache_recorder = cache
+        return instance
 
     def _add_cache_record(self, source, value):
         record = dict(source=source, value=value)
         self._cache_recorder.append(record)
+
+    def render(self):
+        renderer = CLIRenderer(self._cache_recorder)
+        for rendered in renderer.render():
+            yield rendered
 
     def _init_data_(self, ref, factory):
         return ref
@@ -202,8 +214,7 @@ def cache_examples(action):
         header = str(example).replace('_', ' ')
         use._comment_(f'### example: {header} ###')
         action.examples[example](use)
-        cache = use.cache()
-        all_examples.extend(cache)
+        all_examples.extend(use.cache)
     return all_examples
 
 
@@ -215,6 +226,6 @@ def examples(action):
     cached_action = cached_plugin['actions'][action.id]
     cached_examples = cached_action['examples']
 
-    cache_render = CLIRenderer(cached_examples)
-    for rendered in cache_render.render():
+    use = CLIUsage.from_cache(cached_examples)
+    for rendered in use.render():
         yield rendered


q2cli/core/usage.py
```diff
@@ -22,13 +22,25 @@ def __init__(self):
...
+
     def _add_cache_record(self, source, value):
         record = dict(source=source, value=value)
         self._cache_recorder.append(record)
 
+    def render(self):
+        renderer = CLIRenderer(self._cache_recorder)
+        for rendered in renderer.render():
+            yield rendered
+
```
andrewsanchez, line 43: What do you think about returning a string here, as `ArtifactAPIUsage.render` does?

```suggestion
    def render(self):
        renderer = CLIRenderer(self._cache_recorder)
        rendered = renderer.render()
        return '\n'.join(list(rendered))


```

thermokarst: I think we should instead think about inverting the question:

What do you think about updating `ArtifactAPIUsage.render` to return a generator, like we're doing here?

thermokarst: The reason I ask this is because `ArtifactAPIUsage` was the first real driver we implemented, so we didn't have as much experience as we did when we got around to implementing the CLIUsage driver.

andrewsanchez: Sounds good to me, but it's not obvious to me why we would prefer to return a generator over a string. Maybe that will become clear to me soon.

thermokarst: I think its certainly a bit subjective, but, generators are usually a good pick for handling data structures that might require lots of iteration (like these usage examples). This allows the caller to control the flow rate of the data, vs trying to drink from a firehose. If you only need the first few elements, then you can get that using a generator, without incurring the penalty of processing everything.

https://wiki.python.org/moin/Generators
https://stackoverflow.com/a/102632

thermokarst: Also, a string might not be the right data type for every interface. Think q2studio - a potential driver there might render PNG images of the steps.

thermokarst:
I think I'm going to take a stab at refactoring `CLIUsage` along the lines of the `DiagnosticUsage` refactor in https://github.com/qiime2/qiime2/pull/561

thermokarst:
> I think I'm going to take a stab at refactoring

Just kidding.

andrewsanchez:
Everything looks good to me, but can you push a hack commit then revert it to get the workflow to run?  I don't think I can trigger it manually since it hasn't run at all due to the previous draft status.  Tests and everything pass locally, but I'd still like to see the CI pass before merging 🤓 .

I like the class method and the idea of instantiating a stateful driver from some serializable input 🧠 .

thermokarst:
> Everything looks good to me,

Thanks!

> but can you push a hack commit then revert it to get the workflow to run

I can, but you'll be waiting a looonnnggg time - remember, our travis ci acct is disabled due to overuse! ☠️ 

Screen Shot 2021-02-15 at 7 07 40 PM

If travis ci _was_ working for us, no need for a new commit, you could just go into the travis UI and re-run the job.

> and I can't push to thermokarst-forks

Can should try again (but remember, the ci is disabled) - I should have the "allow commits from maintainers" feature on (I never turn that off when I open PRs), but your acct had read-only permissions on this repo, which conflicted with your org-wide permissions. I removed those read-only perms, so in theory you should be set. The one "gotcha" though is that `thermokarst-forks` is an org, not an individual acct, so the "allow commits from maintainers" might not apply here.

> I'd still like to see the CI pass before merging 

Without travis, your best bet is to run locally!

> I like the class method and the idea of instantiating a stateful driver from some serializable input

Thanks! 💝

andrewsanchez (approved):
👌 👏 👏

PR #261, opened 2021-07-08
FEAT: q2cli cast metadata tool

lizgehret:
Casting metadata columns to new q2:types in q2cli should be accomplished via a new `qiime tools cast-metadata` tool. This will allow users to programmatically override the automatically selected q2:types without resorting to tools like `sed`.


q2cli/builtin/tools.py
```diff
@@ -195,7 +195,89 @@ def peek(path):
...
+                    ' Providing multiple file paths to this command will merge'
+                    ' the metadata.',
+               cls=ToolCommand)
+# NOTE: # This gives a potential way to avoid multiple --cast calls but it's
+# fiddely https://stackoverflow.com/a/47730333
+@click.option('--cast', required=True, multiple=True,
+              help='Cast flags for each metadata column that should be'
+              ' specified as either categorical or numeric column types.')
+@click.option('--error-on-extra', is_flag=True,
+              help='Cast flags provided inline that do not exist within the'
```
thermokarst, line 212: This should somehow indicate that we are talking about metadata _columns_ ("cast flags provided inline that do not correspond to columns names in the provided metadata...").

q2cli/builtin/tools.py
```diff
@@ -195,7 +195,89 @@ def peek(path):
...
+              ' metadata provided will result in a raised error.'
+              ' Enabled by default. If this flag is disabled and'
```
thermokarst, line 214: > Enabled by default

I think there might be an error somewhere in here, because there is no obvious way (to me at least) to _disable_ this parameter:

```bash
(q2dev)
~/projects/qiime2/q2cli
☻  cat md1.tsv                                                                                                 cast_metadata_tool 6ccc7c7 ✗
id      foo     bar
a       1       one
b       2       two
c       3       three
(q2dev)
~/projects/qiime2/q2cli
☻  qiime tools cast-metadata md1.tsv --cast foo:categorical spleen:numeric --output-file md1out.tsv            cast_metadata_tool 6ccc7c7 ✗
Usage: qiime tools cast-metadata [OPTIONS] METADATA...
...
                    There was a problem with the command:
 (1/1) Invalid value for 'METADATA...': File 'spleen:numeric' does not exist.
(q2dev)
~/projects/qiime2/q2cli
☻  qiime tools cast-metadata md1.tsv --cast foo:categorical spleen:numeric --output-file md1out.tsv --error-on-extra
Usage: qiime tools cast-metadata [OPTIONS] METADATA...
...
                    There was a problem with the command:
 (1/1) Invalid value for 'METADATA...': File 'spleen:numeric' does not exist.
(q2dev)
~/projects/qiime2/q2cli
☻  qiime tools cast-metadata md1.tsv --cast foo:categorical spleen:numeric --output-file md1out.tsv --error-on-extra=False
Usage: qiime tools cast-metadata [OPTIONS] METADATA...
...
                    There was a problem with the command:
 (1/1?) --error-on-extra option does not take a value
```

q2cli/builtin/tools.py
```diff
@@ -195,7 +195,89 @@ def peek(path):
...
+              ' extra flags are provided, they will be ignored.')
+@click.option('--ignore-missing', is_flag=True,
+              help='Cast flags not provided inline that exist within the'
```
thermokarst, line 217: Same comment as above RE mentioning the column name.

q2cli/builtin/tools.py
```diff
@@ -195,7 +195,89 @@ def peek(path):
...
+              ' metadata provided will be ignored. Enabled by default.'
+              ' If this flag is disabled and not all column flags'
+              ' are provided, an error will be raised.')
+@click.option('--output-file', required=True,
+              type=click.Path(exists=False, file_okay=True, dir_okay=True,
```
thermokarst, line 222: > dir_okay=True

There appears to be a minor logic issue here:

```bash
(q2dev)
~/projects/qiime2/q2cli
☻  mkdir md1out                                                                                                cast_metadata_tool 6ccc7c7 ✗
(q2dev)
~/projects/qiime2/q2cli
☻  qiime tools cast-metadata md1.tsv --cast foo:categorical --output-file md1out/                              cast_metadata_tool 6ccc7c7 ✗
Traceback (most recent call last):
  File "/Users/matthew/.conda/envs/q2dev/bin/qiime", line 33, in <module>
    sys.exit(load_entry_point('q2cli', 'console_scripts', 'qiime')())
  File "/Users/matthew/.conda/envs/q2dev/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/matthew/.conda/envs/q2dev/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/matthew/.conda/envs/q2dev/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/matthew/.conda/envs/q2dev/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/matthew/.conda/envs/q2dev/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/matthew/.conda/envs/q2dev/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/Users/matthew/projects/qiime2/q2cli/q2cli/builtin/tools.py", line 277, in cast_metadata
    cast_md.save(output_file)
  File "/Users/matthew/projects/qiime2/qiime2/qiime2/metadata/metadata.py", line 222, in save
    MetadataWriter(self).write(filepath)
  File "/Users/matthew/projects/qiime2/qiime2/qiime2/metadata/io.py", line 346, in write
    with open(filepath, 'w', newline='', encoding='utf-8') as fh:
IsADirectoryError: [Errno 21] Is a directory: 'md1out/'
```

q2cli/builtin/tools.py
```diff
@@ -195,7 +195,89 @@ def peek(path):
...
+    metadata = _merge_metadata(paths)
+
+    # TODO: Refactor into separate function
+    # Parsing cast arguments into key:value pairs
+    try:
+        cast_dict = {k: v for k, v in (elem.split(':') for elem in cast)}
+    except Exception as err:
+        header = \
+            'Could not parse provided cast arguments into key:value pairs.'
+        ' Please make sure all castes are of the format --cast COLUMN:TYPE'
```
thermokarst, line 243: "castes" -> "cast flags" (?)

q2cli/builtin/tools.py
```diff
@@ -195,7 +195,89 @@ def peek(path):
...
+
+    # Remove entries from the cast dict that are not in the metadata to avoid
+    # errors further down the road
+    for cast in cast_names:
+        if cast not in column_names:
+            cast_dict.pop(cast)
+
+    with tempfile.NamedTemporaryFile() as temp:
+        metadata.save(temp.name)
+        cast_md = qiime2.Metadata.load(temp.name, cast_dict)
```
thermokarst, line 276: Clever! I forgot about that `.load` option to override the types, cool!

thermokarst (changes requested):
Woot! This is looking so great, thanks for the hard work on this you two! I have a few minor observations inline - let me know if you have any questions!

q2cli/builtin/tools.py
lizgehret: 🤓 😏

q2cli/builtin/tools.py
```diff
@@ -195,7 +195,89 @@ def peek(path):
         click.echo(metadata.format)
 
 
-@tools.command('inspect-metadata',
+@tools.command(name='cast-metadata',
+               short_help='Designate metadata columns as numerical'
+                          ' or categorical.',
```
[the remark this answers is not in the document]
lizgehret: Thanks @thermokarst! What are your thoughts on including a link to the Metadata docs page with the supported column types? Something like 'supported column types can be found here: https://docs.qiime2.org/2021.4/tutorials/metadata/?highlight=qiime%20tools#column-types' so that users have an easy reference if they are unsure of the column types they can use.

q2cli/builtin/tools.py
lizgehret: Good catch @thermokarst, this should be set to false. Will update this!

q2cli/builtin/tools.py
lizgehret: @thermokarst thanks for pointing this out! These flags do need to be modified a bit - will get this updated as well.

thermokarst: I would like to see this problem resolved without the introduction of two additional flags to this command - I am concerned that that will make the user-facing API a bit more complicated than it needs to be here. Please take a look at the `--show-importable-types` and `--show-importable-formats` flags on the `qiime tools import` command for some inspiration on solving this using only two flags.

thermokarst: Also, take a look at the docs here for a nice example: https://click.palletsprojects.com/en/8.0.x/options/#boolean-flags

q2cli/builtin/tools.py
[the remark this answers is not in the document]
thermokarst: I think including a link in the `short_help` text will make the prose a bit unreadable, unfortunately.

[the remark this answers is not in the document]
thermokarst: To be clear, I do like that idea in the full help text, although you'll need to sort out how to handle the URL versioning - we don't want to make a maintenance issue here when releasing new version of QIIME 2.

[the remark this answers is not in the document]
lizgehret: @thermokarst a better solution that we pushed up is just including docs.qiime2.org as a general push in the right direction for folks who are unsure. Let me know what you think of that formatting!

q2cli/builtin/tools.py
lizgehret: @thermokarst sounds good - we've updated the logic for these flags, and reverted back to just using two flags in total.

q2cli/builtin/tools.py
```diff
@@ -195,7 +196,98 @@ def peek(path):
...
+            param_hint='cast')
+
+    column_names = set(metadata.columns.keys())
+    cast_names = set(cast_dict.keys())
+
+    if not no_error_on_extra:
+        if not cast_names.issubset(column_names):
+            raise click.BadParameter(
+                message='One or more cast columns were not found within the'
+                ' metadata.',
```
thermokarst, line 268: Telling the caller what the offending columns are would be really helpful. This is a nice-to-have, but given that y'all are pretty close to wrapping this work up, I think you should strongly consider including that now.

q2cli/builtin/tools.py
```diff
@@ -195,7 +196,98 @@ def peek(path):
...
+                param_hint='cast')
+
+    if no_ignore_missing:
+        if not column_names.issubset(cast_names):
+            raise click.BadParameter(
+                message='One or more columns within the metadata'
+                ' were not provided in the cast.',
```
thermokarst, line 275: Same as above, telling the caller what the offending columns are would be really helpful.

thermokarst (changes requested):
I'm loving these tests, great work you two! We are very much down to the nitty-gritty now - please see comments inline. Thanks!

q2cli/builtin/tools.py
```diff
@@ -15,6 +15,7 @@
 
 
 _COMBO_METAVAR = 'ARTIFACT/VISUALIZATION'
+_COMBO_CASTVAR = 'COLUMN:TYPE'
```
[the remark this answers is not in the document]
lizgehret: copy-pasta indeed 😅 
Thanks @thermokarst, putting it inline is a much better idea - will go with that instead since we are only using it once!

q2cli/builtin/tools.py
```diff
@@ -195,7 +196,98 @@ def peek(path):
...
+                    ' the metadata.',
+               cls=ToolCommand)
+@click.option('--cast', required=True, metavar=_COMBO_CASTVAR, multiple=True,
+              help='Cast parameter for each metadata column that should'
+              ' contain a specified column type (supported types can be'
+              ' found on docs.qiime2.org). The required formatting for this'
+              ' parameter is --cast COLUMN:TYPE, repeated for each column'
+              ' and the associated column type it should be cast to in'
+              ' the resultant output.')
+@click.option('--no-error-on-extra', is_flag=True,
```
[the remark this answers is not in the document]
lizgehret: Agreed, good call here - I think we've gone a full 360 degrees on the naming/logic on these flags 😅

q2cli/builtin/tools.py
```diff
@@ -195,7 +196,98 @@ def peek(path):
...
+
+    metadata = _merge_metadata(paths)
+
+    cast_dict = {}
+    try:
+        for casting in cast:
+            k, v = casting.split(':')
+            if k in cast_dict:
+                raise click.BadParameter(
+                    message=(f'Column name "{k}" appears in cast more than'
```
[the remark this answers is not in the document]
lizgehret: Full credit to @Oddant1 on this one! 🥳

q2cli/builtin/tools.py
lizgehret: @thermokarst my only concern with adding this for columns not included in the metadata file is for the scenario that someone only wants to change say, one or two columns in a large metadata file. Do you still think that's reasonable as an output, even if it ends up being a really long list?

q2cli/builtin/tools.py
lizgehret: Agreed, that's a really nice add-on - will get this added!

q2cli/tests/test_tools.py
```diff
@@ -15,10 +15,98 @@
...
+                self.tempdir, 'metadata.tsv')
+        with open(self.metadata_file, 'w') as f:
+            f.write('id\tnumbers\tstrings\n0\t42\tabc\n1\t-1.5\tdef\n')
+
+        self.output_file = os.path.join(
+                self.tempdir, 'test_output.tsv')
+
+    def test_input_invalid_column_type(self):
+        result = self.runner.invoke(
+            tools, ['cast-metadata', self.metadata_file, '--cast', 'id:foo',
```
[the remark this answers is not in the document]
lizgehret: Ohhh good call @thermokarst, thank you - will get this updated in all of our tests!

q2cli/builtin/tools.py
```diff
@@ -195,7 +196,98 @@ def peek(path):
...
+                             ' once.'),
+                    param_hint='cast')
+            cast_dict[k] = v
+    except Exception as err:
+        header = \
+            ('Could not parse provided cast arguments into unique key:value'
```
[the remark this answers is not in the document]
lizgehret: Good call, will update this as well!

q2cli/builtin/tools.py
thermokarst: > the scenario that someone only wants to change say, one or two columns in a large metadata file

If they are doing that, then they wouldn't have the flag enabled, right (which should be the default behavior)? So then this error wouldn't show up for them at all, either way. The point of showing this error is to alert a user who has _opted into_ this behavior. I would show them everything that is wrong, since that's basically what they're asking for.

thermokarst: This is a really good question, btw! Thanks for thinking about these things!

lizgehret: @thermokarst fair enough, that's a good point - I agree with that! I've updated the logging for this scenario to now include all of the missing columns not included in the cast 🙂

q2cli/builtin/tools.py
```diff
@@ -195,7 +196,104 @@ def peek(path):
         click.echo(metadata.format)
 
 
-@tools.command('inspect-metadata',
+@tools.command(name='cast-metadata',
+               short_help='Designate metadata column types.',
+               help='Designate metadata column types.'
+                    ' Supported column types can be found on docs.qiime2.org.'
```
thermokarst, line 202: Awesome, this is shaping up well! One idea instead of putting docs.qiime2.org in here - you could interpolate the values of `_COLUMN_TYPES` into the string here.

q2cli/builtin/tools.py
```diff
@@ -195,7 +196,104 @@ def peek(path):
...
+                    ' Providing multiple file paths to this command will merge'
+                    ' the metadata.',
+               cls=ToolCommand)
+@click.option('--cast', required=True, metavar='COLUMN:TYPE', multiple=True,
+              help='Cast parameter for each metadata column that should'
+              ' contain a specified column type (supported types can be'
+              ' found on docs.qiime2.org). The required formatting for this'
```
thermokarst, line 209: Same note as above RE docs.qiime2.org, if you decide to interpolate.

q2cli/builtin/tools.py
```diff
@@ -195,7 +196,104 @@ def peek(path):
...
+@tools.command(name='cast-metadata',
+               short_help='Designate metadata column types.',
+               help='Designate metadata column types.'
+                    ' Supported column types can be found on docs.qiime2.org.'
+                    ' Providing multiple file paths to this command will merge'
+                    ' the metadata.',
+               cls=ToolCommand)
+@click.option('--cast', required=True, metavar='COLUMN:TYPE', multiple=True,
+              help='Cast parameter for each metadata column that should'
+              ' contain a specified column type (supported types can be'
```
thermokarst, line 208: "Cast" is a tricksy hobbit in prose, since it can be read as a verb here, which is a bummer. How about something like this:

```
              help='Parameter for each metadata column that should be'
              ' cast as a specified column type (supported types can be'
```

q2cli/builtin/tools.py
```diff
@@ -195,7 +196,104 @@ def peek(path):
...
+              ' found on docs.qiime2.org). The required formatting for this'
+              ' parameter is --cast COLUMN:TYPE, repeated for each column'
+              ' and the associated column type it should be cast to in'
+              ' the resultant output.')
```
thermokarst, line 212: Not that we're running low on available space here or anything, but "resultant output" is redundant, let's remove one of these two words: your choice!

q2cli/builtin/tools.py
```diff
@@ -195,7 +196,104 @@ def peek(path):
...
+@click.option('--ignore-extra', is_flag=True,
+              help='Cast parameters that do not correspond to any of the'
+              ' column names within the provided metadata will result in'
+              ' a raised error unless this flag is enabled.')
```
thermokarst, line 216: 🦑

The help text here is mostly describing what happens when the flag _isn't_ provided, which requires a little bit of mental overhead thinking about (thinking is no fun) - can you tweak the verbiage to describe what happens when the flag is enabled, primarily? This is likely an indicator that that root help text for the action is insufficient - moving some of this text up into the root help text should set you up well for succinct explanation here.

q2cli/builtin/tools.py
```diff
@@ -195,7 +196,104 @@ def peek(path):
...
+@click.option('--error-on-missing', is_flag=True,
+              help='If this flag is enabled, failing to include cast'
+              ' parameters for all columns in the provided metadata will'
+              ' result in an error.')
```
thermokarst, line 220: Perfection!

q2cli/builtin/tools.py
```diff
@@ -195,7 +196,104 @@ def peek(path):
...
+              help='Path to file or directory where the'
+              ' modified metdata should be written to.')
+@click.argument('paths', nargs=-1, required=True, metavar='METADATA...',
+                type=click.Path(exists=True, file_okay=True, dir_okay=False,
+                                readable=True))
+def cast_metadata(paths, cast, output_file, ignore_extra,
+                  error_on_missing):
+    import tempfile
+    import qiime2
+    import numpy as np
```
thermokarst, line 233: I don't see a corresponding update listing `numpy` as a dependency in the [package recipe](https://github.com/qiime2/q2cli/blob/master/ci/recipe/meta.yaml), which is an issue in general when adding new package deps.

Bigger picture though, no need for `numpy` here for basic set operations, I'll explain below! Follow me...

q2cli/builtin/tools.py
```diff
@@ -195,7 +196,104 @@ def peek(path):
...
+                     ' type. Valid types: %s' %
+                     (sorted(_COLUMN_TYPES))),
+            param_hint='cast')
+
+    column_names = set(metadata.columns.keys())
+    cast_names = set(cast_dict.keys())
+
+    if not ignore_extra:
+        if not cast_names.issubset(column_names):
+            cast = np.setdiff1d(cast_names, column_names)
```
thermokarst, line 268: This kind of set difference is actually pretty straightforward to handle in python without pulling in something like `numpy`, you just have to know where to look, which is a challenge!

https://docs.python.org/3.8/library/stdtypes.html#set-types-set-frozenset

example:

```python
x = {1, 2, 3, 4}
y = {3, 4, 5, 6}

set_diff = x - y  # or set_diff = x.difference(y)
assert set_diff == {1, 2}
```

q2cli/builtin/tools.py
```diff
@@ -195,7 +196,104 @@ def peek(path):
...
+            raise click.BadParameter(
+                message=('The following cast columns were not found'
+                         ' within the metadata: %s' %
+                         (sorted(cast))),
+                param_hint='cast')
+
+    if error_on_missing:
+        if not column_names.issubset(cast_names):
+            cols = np.setdiff1d(column_names, cast_names)
```
thermokarst, line 277: Same comment as above RE built-in set ops.

q2cli/tests/test_tools.py
```diff
@@ -15,10 +15,97 @@
...
+        self.assertIn('Unknown column type provided.', result.output)
+
+    def test_input_duplicate_columns(self):
+        result = self.runner.invoke(
+            tools, ['cast-metadata', self.metadata_file, '--cast',
+                    'numbers:numerical', '--cast', 'numbers:categorical',
+                    '--output-file', self.output_file])
+
+        self.assertNotEqual(result.exit_code, 0)
+        self.assertIn('appears in cast more than once.', result.output)
```
thermokarst, line 50: Fabu! Can you also check that the offending column name is correctly captured in the error message?

q2cli/tests/test_tools.py
```diff
@@ -15,10 +15,97 @@
...
+            result.output)
+
+    def test_error_on_extra(self):
+        result = self.runner.invoke(
+            tools, ['cast-metadata', self.metadata_file, '--cast',
+                    'extra:numeric', '--output-file', self.output_file])
+
+        self.assertNotEqual(result.exit_code, 0)
+        self.assertIn(
+            'The following cast columns were not found within the metadata:',
```
thermokarst, line 70: I'm on the edge of my seat here.... which columns?!?!

q2cli/tests/test_tools.py
```diff
@@ -15,10 +15,97 @@
...
+    def test_error_on_missing(self):
+        result = self.runner.invoke(
+            tools, ['cast-metadata', self.metadata_file, '--cast',
+                    'numbers:categorical', '--error-on-missing',
+                    '--output-file', self.output_file])
+
+        self.assertNotEqual(result.exit_code, 0)
+        self.assertIn(
+            'The following columns within the metadata'
+            ' were not provided in the cast:',
```
thermokarst, line 82: Same as above!

thermokarst (changes requested):
This is ssssoooo awesome! I'll give this back for another round, then I'll give it one last interactive inspection.

BTW - any thoughts on making `--output-file` optional (and emitting the output to stdout)? No worries either way.

q2cli/builtin/tools.py
```diff
@@ -290,7 +289,13 @@ def cast_metadata(paths, cast, output_file, ignore_extra,
     with tempfile.NamedTemporaryFile() as temp:
         metadata.save(temp.name)
         cast_md = qiime2.Metadata.load(temp.name, cast_dict)
+
+    if output_file:
         cast_md.save(output_file)
+    else:
```
lizgehret, line 295: After doing some testing on the docs rendering, this actually won't work because it's not truly writing to stdout - so this is still a WIP! Working on this today, and will push up some changes after some more testing.

q2cli/builtin/tools.py
```diff
@@ -290,7 +289,13 @@ def cast_metadata(paths, cast, output_file, ignore_extra,
...
+        with tempfile.NamedTemporaryFile(mode='w+') as stdout_temp:
+            cast_md.save(stdout_temp.name)
+            print(stdout_temp.read(), end='')
```
lizgehret, line 298: After doing some testing on the docs rendering, this actually won't work because it's not truly writing to stdout - so this is still a WIP! Working on this today, and will push up some changes after some more testing.

lizgehret: Just pushed up some changes that should resolve this for the docs rendering - will circle back after rendering the docs again to see if that does the trick!

q2cli/builtin/tools.py
```diff
@@ -195,7 +196,111 @@ def peek(path):
...
+        metadata.save(temp.name)
+        cast_md = qiime2.Metadata.load(temp.name, cast_dict)
+
+    if output_file:
+        cast_md.save(output_file)
+    else:
+        with tempfile.NamedTemporaryFile(mode='w+') as stdout_temp:
+            cast_md.save(stdout_temp.name)
+            stdout_str = stdout_temp.read()
+            sys.stdout.write(stdout_str)
```
thermokarst, line 300: Ahhh yes, good catch! I think in this repo we primarily use a click utility (`click.echo`) for handling this kind of stream writing:

https://github.com/qiime2/q2cli/search?q=click.echo

Can you update this to make it consistent with the rest of the package? Thanks!

q2cli/builtin/tools.py
thermokarst: The error message raised by invalid castings is pretty hairy:

```bash
qiime tools cast-metadata --cast bar:numeric md1.tsv
Traceback (most recent call last):
  File "/Users/matthew/projects/qiime2/qiime2/qiime2/metadata/io.py", line 109, in read
    df = df.apply(self._cast_column, axis='index',
  File "/Users/matthew/.conda/envs/q2dev/lib/python3.8/site-packages/pandas/core/frame.py", line 7768, in apply
    return op.get_result()
  File "/Users/matthew/.conda/envs/q2dev/lib/python3.8/site-packages/pandas/core/apply.py", line 185, in get_result
    return self.apply_standard()
  File "/Users/matthew/.conda/envs/q2dev/lib/python3.8/site-packages/pandas/core/apply.py", line 276, in apply_standard
    results, res_index = self.apply_series_generator()
  File "/Users/matthew/.conda/envs/q2dev/lib/python3.8/site-packages/pandas/core/apply.py", line 290, in apply_series_generator
    results[i] = self.f(v)
  File "/Users/matthew/.conda/envs/q2dev/lib/python3.8/site-packages/pandas/core/apply.py", line 110, in f
    return func(x, *args, **kwds)
  File "/Users/matthew/projects/qiime2/qiime2/qiime2/metadata/io.py", line 296, in _cast_column
    return self._to_numeric(series)
  File "/Users/matthew/projects/qiime2/qiime2/qiime2/metadata/io.py", line 326, in _to_numeric
    raise MetadataFileError(
qiime2.metadata.io.MetadataFileError: Cannot convert metadata column 'bar' to numeric. The following values could not be interpreted as numeric: 'one', 'three', 'two'

There may be more errors present in the metadata file. To get a full report, sample/feature metadata files can be validated with Keemei: https://keemei.qiime2.org

Find details on QIIME 2 metadata requirements here: https://docs.qiime2.org/2021.8/tutorials/metadata/

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/matthew/.conda/envs/q2dev/bin/qiime", line 33, in <module>
    sys.exit(load_entry_point('q2cli', 'console_scripts', 'qiime')())
  File "/Users/matthew/.conda/envs/q2dev/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/matthew/.conda/envs/q2dev/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/matthew/.conda/envs/q2dev/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/matthew/.conda/envs/q2dev/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/matthew/.conda/envs/q2dev/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/matthew/.conda/envs/q2dev/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/Users/matthew/projects/qiime2/q2cli/q2cli/builtin/tools.py", line 292, in cast_metadata
    cast_md = qiime2.Metadata.load(temp.name, cast_dict)
  File "/Users/matthew/projects/qiime2/qiime2/qiime2/metadata/metadata.py", line 328, in load
    return MetadataReader(filepath).read(into=cls,
  File "/Users/matthew/projects/qiime2/qiime2/qiime2/metadata/io.py", line 120, in read
    raise MetadataFileError(msg, include_suffix=False)
qiime2.metadata.io.MetadataFileError: Cannot convert metadata column 'bar' to numeric. The following values could not be interpreted as numeric: 'one', 'three', 'two'

There may be more errors present in the metadata file. To get a full report, sample/feature metadata files can be validated with Keemei: https://keemei.qiime2.org

Find details on QIIME 2 metadata requirements here: https://docs.qiime2.org/2021.8/tutorials/metadata/
```

One way to clean this up is to wrap the metadata parsing errors:

```python
        try:
            cast_md = qiime2.Metadata.load(temp.name, cast_dict)
        except qiime2.metadata.io.MetadataFileError as e:
            raise click.BadParameter(message=e, param_hint='cast') from e
```

```bash
qiime tools cast-metadata --cast bar:numeric md1.tsv
Usage: qiime tools cast-metadata [OPTIONS] METADATA...

Error: Invalid value for cast: Cannot convert metadata column 'bar' to numeric. The following values could not be interpreted as numeric: 'one', 'three', 'two'

There may be more errors present in the metadata file. To get a full report, sample/feature metadata files can be validated with Keemei: https://keemei.qiime2.org

Find details on QIIME 2 metadata requirements here: https://docs.qiime2.org/2021.8/tutorials/metadata/
```

q2cli/builtin/tools.py
```diff
@@ -195,7 +196,111 @@ def peek(path):
...
+              ' correspond to any of the column names within the provided'
+              ' metadata will be ignored.')
+@click.option('--error-on-missing', is_flag=True,
+              help='If this flag is enabled, failing to include cast'
+              ' parameters for all columns in the provided metadata will'
+              ' result in an error.')
+@click.option('--output-file', required=False,
+              type=click.Path(exists=False, file_okay=True, dir_okay=False,
+                              writable=True),
+              help='Path to file or directory where the'
```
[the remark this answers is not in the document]
lizgehret: @thermokarst it sure is 😅 thanks for catching this! I'll go ahead and remove that parameter!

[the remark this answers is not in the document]
thermokarst: Oops, I think you removed the wrong thing - I'll follow up in my next round of review.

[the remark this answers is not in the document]
lizgehret: OH duh, now I understand 😅 thanks @thermokarst!

q2cli/builtin/tools.py
```diff
@@ -7,14 +7,15 @@
 # ----------------------------------------------------------------------------
 
 import os
-
+import qiime2
```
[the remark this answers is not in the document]
lizgehret: @thermokarst thanks for bringing this up! I'm wondering if the following modification will make a difference (in terms of import time):

`from qiime2.metadata.base import SUPPORTED_COLUMN_TYPES as _COLUMN_TYPES`

Does this prevent the entire qiime2 import from occurring?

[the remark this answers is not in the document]
thermokarst: >  I'm wondering if the following modification will make a difference (in terms of import time)

This is a great suggestion, let's take some measurements!

Screen Shot 2021-07-20 at 1 25 03 PM

😞 

So to be fair, the framework isn't _that_ slow, the main source for slowness is when loading up an instance of the plugin manager (which you aren't doing here, fortunately). But, the precedent in this module is to defer import of `qiime2` until the last possible moment, because it could be slow.

Give the cache a shot and let me know how it goes - I think once you get your bearings it'll be relatively quick and easy.

q2cli/tests/test_tools.py
```diff
@@ -15,10 +15,109 @@
...
+
+    def test_write_to_stdout(self):
+        result = self.runner.invoke(
+            tools, ['cast-metadata', self.metadata_file, '--cast',
+                    'numbers:categorical'])
+
+        self.assertEqual(result.exit_code, 0)
+        self.assertEqual(self.cast_metadata_dump, result.output)
+
+
```
thermokarst, line 120: Let's add one more test to help us out here if any additional column types get added to the `qiime2` framework:

```python
from qiime2.metadata.base import SUPPORTED_COLUMN_TYPES

def test_valid_column_types(self):
    result = self.runner.invoke(tools, ['cast-metadata', '--help'])
    for col_type in SUPPORTED_COLUMN_TYPES:
        self.assertIn(col_type, result.output)
```

This is untested code, so you'll probably need to tweak it a bit, but this should make it so that this test will fail if a new column type is added (like `time`) - prompting us to go ahead and update the hardcoded column types in `tools.py`.

Also note, we don't care about slow import times nearly as much here in the test suite.

q2cli/commands.py
```diff
@@ -8,13 +8,13 @@
 
 import click
 
+from q2cli.click.command import BaseCommandMixin
+from q2cli.core.config import CONFIG
+
```
thermokarst, line 13: Please revert this, we wound up not needing to make any changes to the import order, so maybe its best to let sleeping dragons lie - thanks!

q2cli/core/cache.py
```diff
@@ -232,7 +232,7 @@ def _get_current_state(self):
         import qiime2.sdk
 
         state = {
-            'plugins': {}
+            'plugins': {},
```
thermokarst, line 235: Please revert this, since we didn't wind up making any meaningful changes to the cache.

q2cli/builtin/tools.py
```diff
@@ -7,13 +7,11 @@
 # ----------------------------------------------------------------------------
 
 import os
-
```
thermokarst, line 10: Please put this blank line back in - it was being used to group std lib imports (like os) from 3rd-party imports (like click).

q2cli/tests/test_tools.py
thermokarst: Screen Shot 2021-07-21 at 2 41 40 PM

Perfect! That is just what I was looking for!

lizgehret: @thermokarst thanks, this is a great addition! Just tested this locally by setting `SUPPORTED_COLUMN_TYPES = {'categorical', 'numeric', 'footype'}` which results in a failure - so this is perfect. Pushed up this added test and those final tweaks to the leftover typos from the cache work, so we should be just about done - let me know if there's anything else we missed!

lizgehret: @thermokarst thanks, this is a great addition! Just tested this locally by setting `SUPPORTED_COLUMN_TYPES = {'categorical', 'numeric', 'footype'}` which results in a failure - so this is perfect. Pushed up this added test and those final tweaks to the leftover typos from the cache work, so we should be just about done - let me know if there's anything else we missed!

thermokarst (approved):
This looks great @lizgehret & @Oddant1! Thanks so much for all your hard work! I have two very minor inline suggestions included here - I am going to apply them and then merge the PR once the CI has passed. Thanks!

PR #273, opened 2022-01-06
Minor tweaks impacting usage

ChrisKeefe:
Closes https://github.com/qiime2/qiime2/issues/612
Closes https://github.com/qiime2/qiime2/issues/613

### DIFF
--- a/q2cli/core/usage.py
+++ b/q2cli/core/usage.py
@@ -160,7 +160,7 @@ class CLIUsage(usage.Usage):
 
     def init_format(self, name, factory, ext=None):
         if ext is not None:
-            name = '%s.%s' % (name, ext)
+            name = '%s.%s' % (name, ext.lstrip('.'))
 
         variable = super().init_format(name, factory, ext=ext)
 
--- a/q2cli/core/usage.py
+++ b/q2cli/core/usage.py
@@ -134,12 +134,12 @@ class CLIUsage(usage.Usage):
         return variable
 
     def import_from_format(self, name, semantic_type,
-                           fmt_variable, view_type=None):
-        variable = super().import_from_format(
-            name, semantic_type, fmt_variable, view_type=view_type)
+                           variable, view_type=None):
+        imported_var = super().import_from_format(
+            name, semantic_type, variable, view_type=view_type)
 
-        in_fp = fmt_variable.to_interface_name()
-        out_fp = variable.to_interface_name()
+        in_fp = variable.to_interface_name()
+        out_fp = imported_var.to_interface_name()
 
         lines = [
             'qiime tools import \\',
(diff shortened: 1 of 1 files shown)


q2cli/core/usage.py
```diff
@@ -134,11 +134,11 @@ def init_artifact(self, name, factory):
...
 
     def import_from_format(self, name, semantic_type,
-                           fmt_variable, view_type=None):
+                           variable, view_type=None):
         variable = super().import_from_format(
-            name, semantic_type, fmt_variable, view_type=view_type)
+            name, semantic_type, variable, view_type=view_type)
 
-        in_fp = fmt_variable.to_interface_name()
+        in_fp = variable.to_interface_name()
```
ebolyen, line 141: There's a namespace collision here with super(), I've seen this handled elsewhere by assigning the result to `var` and the parameter remains `variables`.

ChrisKeefe: Thanks for the catch. Why do we use two separate variables here, when we could just do all of the recording with the original `variable` arg, and then return the result of the `super().etc()` call at the end.

Is using `var` and `variable` about keeping us in line with the "sandwich pattern" mentioned in the dev docs so the method implementations feel similar? Are we concerned that the super method could side-affect this function's arguments? Are we trying to highlight the different-ness of the passed and super-generated UsageVariables?

thermokarst: I don't think we could do all of the recording with the original `variable` - these represent two completely different pieces of data. The `variable` is the raw format data, a `table.biom`, for example. The `var` is an `Artifact`, it represents the imported data.

`UsageVariable` is not a recorder - the `self.recorder` on this interface is responsible for recording state. Hope that helps.

q2cli/core/usage.py
```diff
@@ -134,11 +134,11 @@ def init_artifact(self, name, factory):
...
-                           fmt_variable, view_type=None):
-        variable = super().import_from_format(
-            name, semantic_type, fmt_variable, view_type=view_type)
+                           variable, view_type=None):
+        var = super().import_from_format(
+            name, semantic_type, variable, view_type=view_type)
 
-        in_fp = fmt_variable.to_interface_name()
+        in_fp = variable.to_interface_name()
         out_fp = variable.to_interface_name()
```
ebolyen, line 142: This has to be changed to `var` (hopefully answering the question for why there are two). You can see these used on lines 153-4

ebolyen (changes requested):

q2cli/core/usage.py
ChrisKeefe: Sorry for all the derping around with this @ebolyen  and @thermokarst  - I've been making bad inferences from intermediate states when I should have been reading the original code. Happy New Year! :tada:

ebolyen (approved):
LGTM, ready for merge when tests pass. Thanks @ChrisKeefe!

ebolyen:
oh they already did, sweet!

PR #275, opened 2022-01-20
BUG: fix namespace collision when names differ

ebolyen:
When mutliple `use.view_as_metadata()` where used on the same original variable (with different names provided), a namespace collision occurs. This happened because the variable returned by super() was actually named after the metadata variable, rather than the user supplied variable.

Behavior is still the same as the interface_name is still immediately overwritten.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@ lint:
 test: all
 	QIIMETEST= pytest
 
+# install pytest-xdist plugin for the `-n auto` argument.
 mystery-stew: all
 	MYSTERY_STEW= pytest -k mystery_stew -n auto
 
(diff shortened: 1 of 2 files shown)


thermokarst (approved):

ebolyen:
mystery-stew has passed

Makefile
```diff
@@ -12,6 +12,7 @@ lint:
 test: all
 	QIIMETEST= pytest
 
+# install pytest-xdist plugin for the `-n auto` argument.
```
thermokarst, line 15: https://github.com/qiime2/q2cli/blob/master/ci/recipe/meta.yaml#L31

ebolyen: oh thanks, I dunno why I didn't think to look there first

PR #276, opened 2022-08-19
HACK: fix List[Bool] types (found by mystery-stew)

ebolyen:
Click is really ending up as a mess for us. We've already gutted a ton of the original source by this point. How much more can it take? 

https://github.com/pallets/click/pull/2293

### DIFF
--- a/q2cli/click/option.py
+++ b/q2cli/click/option.py
@@ -54,7 +54,18 @@ class GeneratedOption(click.Option):
         else:
             attrs['type'] = None
 
+        # This nonsense:
+        # https://github.com/pallets/click/blob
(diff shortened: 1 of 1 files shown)


ebolyen:
Looked into this a bit more, the edits to community comments in the initial change by maintainers in https://github.com/pallets/click/pull/2248 are a little surprising to me. They don't appear disrespectful, but do note that the breaking change was unusual for a point release, which is generally true, although perhaps Pallet's projects don't follow the usual convention (we certainly don't). Still, I find it unsettling to edit out their specific objections.

ebolyen:
If click explodes, it's my fault.

PR #277, opened 2022-08-31
NEW: First pass at cli for cache

Oddant1:
Start adding cli support for the cache so it can enter wider use

### DIFF
--- a/q2cli/util.py
+++ b/q2cli/util.py
@@ -7,7 +7,7 @@
 # ----------------------------------------------------------------------------
 
 
-class ControlFlowException(Exception):
+class OutOfDisk(Exception):
     pass
 
 
@@ -74,7 +74,10 @@ def exit_with_error(e, header='An error has been encountered:',
     if not footer:
         click.echo(err=True)  # extra newline to look normal
 
-    click.get_current_context().exit(status)
+    try:
+        click.get_current_context().exit(status)
+    except RuntimeError:
+        sys.exit(status)
 
 
 def output_in_cache(fp):
@@ -148,10 +151,12 @@ class pretty_failure:
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        if exc_val is not None:
+        # if exit_with_error is called twice, then click.exit(1) or sys.exit(1)
+        # will happen, no need to exit_with_error again in that case.
+        if exc_val is not None and str(exc_val) != '1':
             exit_with_error(exc_val, self.header, self.traceback, self.status)
 
-        return True
+        return False
 
 
 def convert_primitive(ast):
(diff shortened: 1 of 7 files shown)


Oddant1:
Regarding [b3a9f8b](https://github.com/qiime2/q2cli/pull/277/commits/b3a9f8b8aca0040c4a110fc5aec1452202a70155), this, unfortunately, loses us the pretty print when the path ends up not existing. If we want it back, we could break our API here and resolve a path within the cache's data directory to hand into this function. This would break our API, but it would be us breaking it internally so *shrug*. Not sure if we want to play with that though.

ebolyen:
@Oddant1, what if we split the difference, if a path contains a colon, we presume it's probably a cache, and we attempt to resolve it as such. If it doesn't resolve, then we treat is at a file (on the off-chance the user hates things that make sense), and if it's not a file but did have the colon, then we raise an error about the key not existing in the cache (which is probably more true than it being a missing file).

Then, if something never had the colon, it continues to produce the file-doesn't-exist error.

(If I'm understanding the current situation correctly)

Oddant1:
@ebolyen, the reason we lost the pretty print is that I got rid of the need for the file to exist in the click.Path argument because if they passed in a cache:key then a file with that path will. . . hopefully not exist. I can still do the logic the way you described, but we won't get the click pretty print from the file not existing because that processing is happening after inspect_metadata is called. There doesn't seem to be a good place to apply that logic before calling inspect_metadata

Oddant1:
Blocked by qiime2/qiime2#645

Oddant1:
@ebolyen, it would be really nice if we could unify how we load inputs and metadata a bit more because there is a whole lot of overlap, but they don't quite use the same machinery. You'll see what I mean if you look at load_metadata and _convert_input. May be out of the scope of this PR, but I already refactored a good chunk of this.

Oddant1:
@ebolyen are you interested in having cd278e1 in this PR? If not, I can hold off on implementing it until this PR is merged. If you're fine with it, I can write up a unit test or two (I manually tested it by messing with my plugin manager)

ebolyen:
Thanks @Oddant1! I think the commit looks great. I'll pull this all down soon and play around.

q2cli/util.py
```diff
@@ -73,6 +77,43 @@ def exit_with_error(e, header='An error has been encountered:',
...
+
+    # Tells us right away this isn't in a cache
+    if ':' not in fp:
+        return False
+
+    split_path = fp.split(':')
+
+    # Account for potential for : in the path
+    cache_path = ':'.join(elem for elem in split_path[:-1])
+    key = split_path[-1]
```
ebolyen, line 95: An easier way to do this and the split is:
```python
cache_path, key = fp.rsplit(':', 1)
```
Which will split from the right, at most once. The unpacking should also always succeed as we have proved there is at least 1 colon in the path.

q2cli/util.py
```diff
@@ -73,6 +77,43 @@ def exit_with_error(e, header='An error has been encountered:',
...
+
+    try:
+        if Cache.is_cache(Path(cache_path)):
```
ebolyen, line 98: We should probably make `Cache.is_cache` indifferent to the input type. Accepting either `str` or `Path`

q2cli/util.py
```diff
@@ -73,6 +77,43 @@ def exit_with_error(e, header='An error has been encountered:',
...
+                    'rules may be found here https://www.askpython.com/python/'
+                    'python-identifiers-rules-best-practices')
+            else:
+                return True
+    except FileNotFoundError as e:
+        # If cache_path doesn't exist, don't treat this as a cache output
+        if 'No such file or directory' in str(e):
+            pass
+        else:
+            raise e
```
ebolyen, line 111: Is this branch reachable given the specific `FileNotFoundError` filter?

q2cli/util.py
```diff
@@ -234,3 +275,119 @@ def get_plugin_manager():
...
+                raise ValueError(f"The path {fp.split(':')[0]} is not a valid"
+                                 " cache") from e
+            else:
+                raise ValueError(f'{fp!r} is not a valid filepath') from e
+        else:
+            raise ControlFlowException(
+                '%r is not a QIIME 2 Artifact (.qza)' % fp) from e
+    # If we get here, all we really know is we failed to get a Result
+    except Exception as e:
+        raise ControlFlowException(
```
ebolyen, line 368: This class name will leak to the user, which probably isn't ideal.

q2cli/util.py
```diff
@@ -234,3 +275,119 @@ def get_plugin_manager():
...
+    """
+    import os
+    from pathlib import Path
+    from qiime2.core.cache import Cache
+
+    split_path = fp.split(':')
+
+    # Handle the potential for : in the cache path
+    cache_path = ':'.join(elem for elem in split_path[:-1])
+    key = split_path[-1]
```
ebolyen, line 385: I would create a helper using the aforementioned `rsplit` method to factor this shared bit out.

q2cli/util.py
```diff
@@ -234,3 +275,119 @@ def get_plugin_manager():
...
+        # before the ':' was not an existing cache
+        if ':' not in fp or artifact is None:
+            artifact = qiime2.sdk.Result.load(fp)
+    except OSError as e:
+        if e.errno == 28:
+            temp = tempfile.tempdir
+            raise ValueError(f'There was not enough space left on {temp!r} '
+                             f'to extract the artifact {fp!r}. (Try '
+                             'setting $TMPDIR to a directory with more '
+                             f'space, or increasing the size of {temp!r})')
```
ebolyen, line 347: We should migrate this to `Result.load()` it would be useful generally.

q2cli/util.py
```diff
@@ -234,3 +275,119 @@ def get_plugin_manager():
...
+
+    get_plugin_manager()
+    try:
+        if ':' in fp:
+            artifact = convert_to_cache_input(fp)
+
+        # If we get here we either had a path without a ':' or we got
+        # None from convert_to_cache_input meaning the part of value
+        # before the ':' was not an existing cache
+        if ':' not in fp or artifact is None:
```
ebolyen, line 339: A more idiomatic way to do this is to initialize `artifact = None` before your first `if ':' in fp:`, then you can make this line:

```python
if artifact is None:
```

q2cli/util.py
```diff
@@ -234,3 +275,119 @@ def get_plugin_manager():
...
+                '%r is not a QIIME 2 Artifact (.qza)' % fp) from e
+    # If we get here, all we really know is we failed to get a Result
+    except Exception as e:
+        raise ControlFlowException(
+            'There was a problem loading %s as a QIIME 2 Result: ' % fp) from e
+
+    return artifact
+
+
+def convert_to_cache_input(fp):
```
ebolyen, line 374: Maybe this should be `try_as_cache_input`?

q2cli/util.py
```diff
@@ -234,3 +275,119 @@ def get_plugin_manager():
...
+
+def get_input(fp):
+    """ Gets a Result from a filepath if possible
+    """
+    import tempfile
+    import qiime2
+    import qiime2.sdk
+
+    get_plugin_manager()
+    try:
```
ebolyen, line 332: I think there may be one too many try's here, or at least I'm struggling to follow what's expected to happen in different situations.

I think this may need some refactoring, but I'm not really sure how to go about it either. It seems like there must be phases we can distill?

q2cli/util.py
```diff
@@ -234,3 +275,119 @@ def get_plugin_manager():
...
+        try:
+            return qiime2.Metadata.load(fp)
+        except Exception as e:
+            # Handle the case where we try to load a raw .qza as Metadata
+            # because we were missing the pugin(s) necessary to load it as an
+            # Artifact
+            if "'utf-8' codec can't decode byte" in str(e) and \
+                    qiime2.sdk.result.Result.peek(fp):
+                e = ValueError("It looks like you have an Artifact but are "
+                               "missing the plugin(s) necessary to load it.")
```
ebolyen, line 299: This kind of circles around the original problem. Perhaps we refactor `get_input` to return more than just the result? And then call-sites can error as they need/want to?

Exceptions are technically something you can pass around as any other object.

ebolyen (changes requested):
Thanks @Oddant1! I think all of the changes to the q2cli guts are great, but these utilities are still a bit mind-bending. It's probably worth trying to rewrite them from scratch, I think they are just kind of tired after so much use-case accretion.

q2cli/util.py
Oddant1: Turns out I already am doing that. I already caste it to a pathlib.Path at the beginning of is_cache.

q2cli/util.py
Oddant1: Probably not. I wanted to be certain that if some other error did somehow get raised like that we didn't just squash it. Should I just remove the else?

q2cli/util.py
Oddant1: The try here is handling all the errors with loading an Artifact that we understand. If we get an issue that we don't understand we raise a ControlFlowException that can be handled by the call site. The reason the call sites also have try's is so _convert_input can pretty print the failure if there is an exception and so load_metadata can try to load the file as metadata instead of as an artifact if the error it got was a ControlFlowException (any error loading the artifact that we don't understand).

q2cli/util.py
Oddant1: In Result.load or deeper down in mount or something where we actually open things?

q2cli/util.py
Oddant1: In theory, it could but it won't with either of the current call sites. I'm not sure I understand why it's a problem if it leaks anyway.

q2cli/util.py
Oddant1: What do we gain from doing that though? Maybe I've just looked at this too much, so it makes too much sense to me now, but in get_input we raise a ControlFlowException if something we didn't anticipate happens, and we raise more informative errors if we do understand what happened. Why would we pass them around instead of just raising them and letting the call site handle them? What do we gain by returning them or something instead?

ebolyen (approved):
Ok, I've thrown the kitchen sink at this PR by this point. Everything seems to be working!

PR #278, opened 2022-11-08
NEW: Start adding cache tools to the cli

Oddant1:
Adds cache utilities to `qiime tools cache-` on the cli.

### DIFF
--- a/q2cli/builtin/tools.py
+++ b/q2cli/builtin/tools.py
@@ -596,10 +596,10 @@ def cache_create(path):
     try:
         Cache(path)
     except Exception as e:
-        header = 'There was a problem creating a cache at %s:' % path
+        header = "There was a problem creating a cache at '%s':" % path
         q2cli.util.exit_with_error(e, header=header, traceback=None)
 
-    success = 'Created cache at %s' % path
+    success = "Created cache at '%s'" % path
     click.echo(CONFIG.cfg_style('success', success))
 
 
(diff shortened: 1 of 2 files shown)


Oddant1:
Blocked by qiime2/qiime2#663

q2cli/builtin/tools.py
```diff
@@ -604,3 +604,247 @@ def citations(path):
...
+            for data in cache.get_data():
+                try:
+                    click.echo(
+                        CONFIG.cfg_style('success', 'Validating: %s' % data))
+                    if not is_uuid4(data):
+                        raise ValueError(
+                            "Item in data directory '%s' is not a valid "
+                            "uuid4." % data)
+
+                    art = Artifact.load(cache.data / data)
```
ebolyen, line 835: I thought we had made it so that this wasn't possible without going through the `Cache` object.

I would have expected something like `a = Cache.load(...); a.validate()`.

Perhaps we should have a `Cache.validate()` method instead?

q2cli/builtin/tools.py
```diff
@@ -604,3 +604,247 @@ def citations(path):
...
+              help='The key to save the artifact under (must be a valid '
+                   'Python identifier).')
+def cache_save(cache_path, artifact_path, key):
+    from qiime2 import Artifact
+    from qiime2.core.cache import Cache
+    from q2cli.core.config import CONFIG
+
+    try:
+        artifact = Artifact.load(artifact_path)
+        cache = Cache(cache_path)
```
ebolyen, line 705: If we `with` this cache before loading the artifact, then there will be fewer file operations.

ebolyen (commented):
I'm going to merge this now so that we can get some early feedback from others on it, but I have some comments for future changes inline.

Also, I think we should standardize the `--path`/`--cache-path` in all commands to be the same thing (probably `--cache` so that save/load make sense?).

PR #280, opened 2022-12-01
IMP: allow qiime tools peek to take multiple arguments, add tsv flag

colinvwood:


lizgehret:
Hey @colinvwood - just finished testing this locally, and it looks like there's one edge case that still needs some tweaks. If I'm running `qiime tools peek` for multiple files that live in another directory from my cwd, here's what the output looks like:
Screen Shot 2022-12-07 at 9 31 47 AM

This should be modified so that the results are the same as when I'm running this command for files inside my cwd:
Screen Shot 2022-12-07 at 9 32 03 AM

colinvwood:
I changed getting the basename of the file path at the last minute because I didn't know if the full path or only the filename was desired, and apparently that messed up the formatting.

lizgehret (approved):
lgtm, thanks @colinvwood!

gregcaporaso:
Just tested this and it works great. Thanks for this contribution @colinvwood and for reviewing @lizgehret!

PR #283, opened 2022-12-12
SQUASH: Need to with in the cache for validate b/c we load artifacts w…

Oddant1:
The cache needs to be withed in on validate because we proceed to load artifacts which uses get_cache, so we need the correct cache set.

### DIFF
--- a/q2cli/builtin/tools.py
+++ b/q2cli/builtin/tools.py
@@ -822,25 +822,28 @@ def cache_validate(path):
 
     try:
         cache = Cache(path)
-        with cache.lock:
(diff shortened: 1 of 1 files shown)


Oddant1:
Maybe we actually want withing in a cache's lock to with in the cache? Maybe make it follow similar rules to withing in a named pool?

ebolyen:
I think needing to use the lock directly suggests that this should be a method within a cache.

Depending on timing tomorrow, that might be a better approach.

Oddant1:
Blocked by qiime2/qiime2#668

PR #284, opened 2022-12-14
View in cache

Oddant1:
Allows for a cache:key to be handed to `qiime tools view`. Also allows visualizations to be saved to cache using `qiime tools cache-save`

### DIFF
--- a/q2cli/builtin/tools.py
+++ b/q2cli/builtin/tools.py
@@ -696,12 +696,18 @@ def cache_garbage_collection(path):
               help='The key to save the artifact under (must be a valid '
                    'Python identifier).')
 def cache_save(cache_path, artifact_path, key):
-    from qiime2 import Artifact
+    from qiime2 import Artifact, Visualization
     from qiime2.core.cache import Cache
     from q2cli.core.config import CONFIG
 
     try:
-        artifact = Artifact.load(artifact_path)
+        try:
+            artifact = Artifact.load(artifact_path)
+        except TypeError as e:
+            if 'Attempting to load Visualization' in str(e):
+                artifact = Visualization.load(artifact_path)
+            else:
+                raise e
         cache = Cache(cache_path)
         cache.save(artifact, key)
     except Exception as e:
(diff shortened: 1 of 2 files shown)


Oddant1:
Just makes _load_input do the heavy lifting for handing a cache:key to `qiime tools view` but skips getting the plugin manager because that is slow and not necessary. If we're cool with this maybe we get rid of the _ in the function name, or maybe we just don't care. Also I'm not seeing any tests for `qiime tools view`.

lizgehret (approved):
seems reasonable to me 🤷‍♀️

q2cli/builtin/tools.py
```diff
@@ -696,12 +686,18 @@ def cache_garbage_collection(path):
...
 
     try:
-        artifact = Artifact.load(artifact_path)
+        try:
+            artifact = Artifact.load(artifact_path)
+        except TypeError as e:
+            if 'Attempting to load Visualization' in str(e):
+                artifact = Visualization.load(artifact_path)
+            else:
+                raise e
```
ebolyen, line 700: `Result.load()` will return a `result` which is either artifact or visualization letting you delete the try-except

q2cli/builtin/tools.py
```diff
@@ -450,20 +450,10 @@ def view(visualization_path, index_extension):
...
-    try:
-        visualization = qiime2.sdk.Visualization.load(visualization_path)
-    # TODO: currently a KeyError is raised if a zipped file that is not a
-    # QIIME 2 result is passed. This should be handled better by the framework.
-    except (zipfile.BadZipFile, KeyError, TypeError):
-        raise click.BadParameter(
-            '%s is not a QIIME 2 Visualization. Only QIIME 2 Visualizations '
-            'can be viewed.' % visualization_path)
 
+    visualization = _load_input(visualization_path, view=True)[0]
```
ebolyen, line 456: I think we still need to assert that `isinstance(visualization, qiime2.Visualization)` is true, and otherwise raise the original error.

ebolyen (changes requested):

q2cli/builtin/tools.py
Oddant1: Oh yeah good catch

ebolyen (approved):
Seems to work well! Thanks @Oddant1!

PR #287, opened 2022-12-16
SQUASH: Rename tools and args

Oddant1:
Renames cache-save/load to cache-store/fetch and standardizes to --cache-path


ebolyen (approved):
Thanks @Oddant1!

Issue #285, opened 2022-12-15
stdout on creation of artifact cache entry


gregcaporaso:
One little thing that would be nice is if the output summary from q2cli was a little different for cache entries that were created, relative to artifacts created as files. For example, I think (current behavior): 

```
Saved FeatureTable[Frequency] to: /projects/microbiome/jgc-cache/:test_table
Saved FeatureData[Sequence] to: ./test_seqs.qza
Saved SampleData[DADA2Stats] to: /projects/microbiome/jgc-cache/:test_stats
```

could be clearer as something like (proposed behavior):

```
Added FeatureTable[Frequency] to cache /projects/microbiome/jgc-cache/ as test_table
Saved FeatureData[Sequence] to: ./test_seqs.qza
Added SampleData[DADA2Stats] to cache /projects/microbiome/jgc-cache/ as test_stats
```

Oddant1:
I believe this was closed by #290

gregcaporaso:
Thanks!

PR #290, opened 2023-03-03
IMP: differentiate stdout between cache entry and file output

colinvwood:
Changed stdout on action completion to differentiate between artifacts added to a cache and artifacts saved to a standard file. 

See #285.


Oddant1 (approved):
This looks good to me. Just needs to conflicts resolved.

PR #291, opened 2023-03-07
FEAT: qiime tools show-types and qiime tools show-formats commands

colinvwood:
Added two commands to show the available semantic types and importable and exportable formats for a qiime deployment. 

Will replace the functionality of the `qiime tools import --show-importable-types` and `qiime tools import --show-importable-formats` commands.

Additionally:
- descriptions are given of each semantic type or format when available
- semantic types or formats can be subsetted to only those of interest 
- a `--tsv` flag is available

Motivated by #289.

### DIFF
--- a/q2cli/builtin/tools.py
+++ b/q2cli/builtin/tools.py
@@ -151,7 +151,7 @@ def _get_matches(words, possibilities, strict=False):
 def show_types(queries, strict, tsv):
     pm = q2cli.util.get_plugin_manager()
 
-    if queries:
+    if len(queries) > 0:
         matches = _get_matches(queries, list(pm.artifact_classes), strict)
     else:
         matches = sorted(list(pm.artifact_classes))
@@ -191,7 +191,7 @@ def show_formats(queries, importable, exportable, strict, tsv):
     portable_formats = pm.importable_formats if importable \
         else pm.exportable_formats
 
-    if queries:
+    if len(queries) > 0:
         matches = _get_matches(queries, portable_formats.keys(), strict)
     else:
         matches = sorted(portable_formats.keys())
--- a/q2cli/builtin/tools.py
+++ b/q2cli/builtin/tools.py
@@ -80,18 +80,19 @@ def export_data(input_path, output_path, output_format):
     click.echo(CONFIG.cfg_style('success', success))
 
 
-def print_importables(importables, tsv):
+
+def print_descriptions(descriptions, tsv):
     if tsv:
-        for importable, description in importables.items():
-            click.echo(f"{importable}\t", nl=False)
+        for value, description in descriptions.items():
+            click.echo(f"{value}\t", nl=False)
             if description:
                 click.echo(description)
             else:
-                click.echo("No description")
+                click.echo()
     else:
         import textwrap
-        for importable, description in importables.items():
-            click.secho(importable, bold=True)
+        for value, description in descriptions.items():
+            click.secho(value, bold=True)
             if description:
                 tabsize = 8
                 wrapped_description = textwrap.wrap(description, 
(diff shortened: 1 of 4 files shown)


gregcaporaso:
Hey @colinvwood, I haven't reviewed this in any detail, but I think it's worth removing the `--show-importable-*` functionality from `qiime tools import ...` in this PR too. I think the best way to do this is, if the user provides one of those flags, print a note saying that that functionality is now accessible instead through the new command and then exit (without printing the types/formats).

q2cli/builtin/tools.py
```diff
@@ -80,40 +80,149 @@ def export_data(input_path, output_path, output_format):
...
+
+
+def deformat_description(description):
+    import re
+    deformatted = re.sub(r"[\t\n]+", ' ', description)
+    despaced = re.sub(r" +", ' ', deformatted)
+    return despaced
+
+
+def get_matches(words, possibilities, cutoff=0.5):
```
gregcaporaso, line 117: How'd you come up with 0.5 as the default? It feels a little low to me, and on a few tests it seems pretty permissive, e.g.:

```bash
$ qiime tools show-types mpledata
ImmutableMetadata
	Immutable sample or feature metadata.

SampleData[AlphaDiversity]
	Alpha diversity values, each associated with a single sample
	identifier.
...
```

The match to `ImmutableMetadata` seems pretty distant. 

I don't have a good feel for what it should be, so mostly I'm wondering if you had a specific motivation for 0.5.

q2cli/builtin/tools.py
```diff
@@ -80,40 +80,149 @@ def export_data(input_path, output_path, output_format):
...
+    from difflib import get_close_matches
+    matches = []
+    for word in words:
+        matches += get_close_matches(word,
+                                     possibilities,
+                                     n=len(possibilities),
+                                     cutoff=cutoff)
+        # substring search
```
gregcaporaso, line 125: Is this substring search not already handled by difflib? I notice that my search example above for `mpledata` matches to `SampleData[JoinedSequencesWithQuality]`, which is a short substring for the type being matched.

q2cli/builtin/tools.py
```diff
@@ -80,40 +80,149 @@ def export_data(input_path, output_path, output_format):
...
+                click.echo(deformat_description(description))
+            else:
+                click.echo()
+    else:
+        import textwrap
+        for value, description in descriptions.items():
+            click.secho(value, bold=True)
+            if description:
+                description = deformat_description(description)
+                tabsize = 8
```
gregcaporaso, line 97: Pull this outside of your for loop, so it's not set every time through.

q2cli/builtin/tools.py
```diff
@@ -80,40 +80,149 @@ def export_data(input_path, output_path, output_format):
...
+                                                    tabsize=tabsize)
+                for line in wrapped_description:
+                    click.echo(f"{line}")
+            else:
+                click.secho("\tNo description", italic=True)
+            click.echo()
+
+
+def deformat_description(description):
+    import re
```
gregcaporaso, line 111: This import happens every time you call this function. I'm not sure if Python recognizes that and doesn't do the import after the first time, or if it's doing work to import on every call. You can probably instead throw this import in at the top of `print_descriptions`, since you're almost certainly going to call `deformat_description` from their.

q2cli/builtin/tools.py
```diff
@@ -80,40 +80,149 @@ def export_data(input_path, output_path, output_format):
...
+)
+@click.argument('formats', nargs=-1)
+@click.option('--importable', is_flag=True,
+              help='List the importable formats.')
+@click.option('--exportable', is_flag=True,
+              help='List the exportable formats.')
+@click.option('--strict', is_flag=True,
+              help='Show only exact matches for the format argument(s).')
+@click.option('--tsv', is_flag=True,
+              help='Print as machine readable tab separated values.')
```
gregcaporaso, line 178: `tab separated` -> `tab-separated`

q2cli/builtin/tools.py
```diff
@@ -80,40 +80,149 @@ def export_data(input_path, output_path, output_format):
...
+        raise click.UsageError("'--importable' and '--exportable' flags are "
+                               "mutually exclusive.")
+    if not importable and not exportable:
+        raise click.UsageError("One of '--importable' or '--exportable' flags "
+                               "is required.")
+
+    pm = q2cli.util.get_plugin_manager()
+    portable_formats = pm.importable_formats if importable \
+        else pm.exportable_formats
+    available_formats = list(portable_formats)
```
gregcaporaso, line 190: It looks like you're defining a second variable here so that you have this information in a list (`available_formats`) and a dict (`portable_formats`). It's a little confusing to read though, as it's not immediately clear why you need need both of these. I'm guessing it's to provide a list to `get_matches`. If that's correct, calling:

```
 get_matches(formats, portable_formats.keys(), ...)
```

would be a little clearer in my opinion.

q2cli/builtin/tools.py
```diff
@@ -80,40 +80,149 @@ def export_data(input_path, output_path, output_format):
...
+    print_descriptions(descriptions, tsv)
+
+
+@tools.command(
+        name='show-formats',
+        help='List the availabe formats.',
+        short_help='',
+        cls=ToolCommand
+)
+@click.argument('formats', nargs=-1)
```
gregcaporaso, line 170: I think these two new functions would be a little easier to follow if `formats` and `types` were called `queries`.

q2cli/builtin/tools.py
```diff
@@ -80,40 +80,149 @@ def export_data(input_path, output_path, output_format):
...
+)
+@click.argument('types', nargs=-1)
+@click.option('--strict', is_flag=True,
+              help='Show only exact matches for the type argument(s).')
+@click.option('--tsv', is_flag=True,
+              help='Print as machine readable tab separated values.')
+def show_types(types, strict, tsv):
+    pm = q2cli.util.get_plugin_manager()
+
+    if types and strict:
```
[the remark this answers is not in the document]
gregcaporaso: I also recommend being more explicit with your test, for readability. In this case, `if len(matches) < 1`.

q2cli/builtin/tools.py
```diff
@@ -80,40 +80,149 @@ def export_data(input_path, output_path, output_format):
...
+    if formats and strict:
+        matches = get_matches(formats, available_formats, 1)
+    elif formats:
+        matches = get_matches(formats, available_formats)
+    else:
+        matches = available_formats
 
-    import q2cli.util
+    descriptions = {}
+    for match in sorted(matches):
```
gregcaporaso, line 200: A side-effect of sorting here is that you lose the sorting that `get_close_matches` provides, which is sorting by decreasing similarity. That can lead to unusual cases like the example I provided above, where the least similar match (`ImmutableMetadata`) was presented first. 

I think you should leave sorting as-is when calling `get_matches`, and sort alphabetically like you're doing here in the case where `len(formats) == 0`.

q2cli/tests/test_tools.py
```diff
@@ -724,5 +725,154 @@ def test_multiple_file_tsv(self):
...
+    def test_show_formats_strict(self):
+        formats = list(self.pm.exportable_formats)[:5]
+        result = self.runner.invoke(tools, ['show-formats', '--exportable',
+                                            '--strict', *formats])
+        self.assertEqual(result.exit_code, 0)
+        print(formats)
+        print(result.output)
+        self.assertEqual(len(result.output.split('\n\n')) - 1, len(formats))
+
+        result = self.runner.invoke(tools, ['show-formats', '--exportable',
```
gregcaporaso, line 840: I think this actually isn't testing that `strict` is working here, because these queries wouldn't return a hit in either case: 

```bash
$ qiime tools show-formats --importable nonsense
$ qiime tools show-formats --importable nonsense --strict
$
```

I would instead create something that is a close match to an existing type, and then confirm that you get expected results here with and without `strict`, to test that toggling the parameter behaves as expected.

gregcaporaso (changes requested):
Hey @colinvwood, I have some requests for changes on this PR. It might look like a lot, but they're all pretty minor so I think this should be relatively quick to address. Let me know if you have any questions.

q2cli/builtin/tools.py
```diff
@@ -80,40 +80,149 @@ def export_data(input_path, output_path, output_format):
...
+            for possibility in possibilities:
+                if word.lower() in possibility.lower():
+                    if possibility not in matches:
+                        matches.append(possibility)
+
+    return matches
+
+
+@tools.command(
+        name='show-types',
```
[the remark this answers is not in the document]
colinvwood: Because that's what the old commands are called

q2cli/builtin/tools.py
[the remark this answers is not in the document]
colinvwood: I don't follow the second point, you mean the number of command line arguments given?

q2cli/builtin/tools.py
colinvwood: No specific motivation, having results that a user is not interested in seemed preferable to omitting results a user is interested in. I'll change to 0.6.

q2cli/builtin/tools.py
colinvwood: No it isn't, only overall similarity based.

colinvwood: It may match because of the explicit substring search, or they may be similar enough

q2cli/builtin/tools.py
colinvwood: The overhead is so tiny (a dictionary miss) that it seems worth it for clarity's sake to keep it in the function that has the dependency, but I'm happy to move it too.

colinvwood:
Not sure why the cache test is failing, maybe @Oddant1 can weigh in?

q2cli/builtin/tools.py
```diff
@@ -767,13 +876,11 @@ def cache_status(cache):
...
 
-                if 'data' in key_values:
-                    data = key_values['data']
+                if (data := key_values['data']) is not None:
                     data_output.append(
                         'data: %s -> %s' %
                         (key, str(Result.peek(_cache.data / data))))
-                elif 'pool' in key_values:
-                    pool = key_values['pool']
+                elif (pool := key_values['pool']) is not None:
```
Oddant1, line 883: @colinvwood Looks like the issue is right here probably from the merge. We want the version where it checks `if 'data' in key_values` and the same for 'pool'

Oddant1: Alternatively, if we want to be cool

```
                if (data := key_values.get('data', None)) is not None:
                    data_output.append(
                        'data: %s -> %s' %
                        (key, str(Result.peek(_cache.data / data))))
                elif (pool := key_values.get('pool', None)) is not None:
                    pool_output.append(
                        'pool: %s -> size = %s' %
                        (key, str(len(os.listdir(_cache.pools / pool)))))
 ```

q2cli/builtin/tools.py
gregcaporaso: I agree that this is an acceptable performance hit for the sake of code clarity.

gregcaporaso (approved):
Looks good to me, thanks for the updates @colinvwood!

Oddant1 (approved):
lgtm

PR #294, opened 2023-04-19
Parsl

Oddant1:
Adds support for parsl features to the cli

### DIFF
--- a/q2cli/commands.py
+++ b/q2cli/commands.py
@@ -229,7 +229,21 @@ class ActionCommand(BaseCommandMixin, click.Command):
                          help='Indicate that you want to execute your action '
                               'with parsl. This flag will check in the '
                               'following pre-defined locations for a parsl '
-                              'config:\n\n1. It will'),
+                              'config:'
+                              '\n\n1. It will check for a file at the '
+                              'filepath set by the QIIME2_CONFIG environment '
+                              'variable.'
+                              '\n\n2. It will check for a file at '
+                              'appdirs.user_config_dir(qiime2)/'
+                              'qiime2_config.toml.'
+                              '\n\n3. It will check for a file at '
+                              'appdirs.site_config_dir(qiime2)/'
+                              'qiime2_config.toml.'
+                              '\n\n4. It will check for a file at '
+                              '$CONDA_PREIFX/etc/qiime2_config.toml'
+                              '\n\n5. If none of the others are found, it '
+                              'will write the default config to $CONDA_PREFIX/'
+                              'etc/qiime2_config.toml and use that.'),
             click.Option(['--parsl-config'], required=False,
                          type=click.Path(exists=True, dir_okay=False),
                          help='Indicate that you want to execute your action '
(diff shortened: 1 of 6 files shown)


Oddant1:
blocked by qiime2/qiime2#635

q2cli/commands.py
```diff
@@ -225,6 +225,17 @@ def __init__(self, name, plugin, action):
                               'during execution of this action. Or silence '
                               'output if execution is successful (silence is '
                               'golden).'),
+            click.Option(['--parsl'], is_flag=True, required=False,
+                         help='Indicate that you want to execute your action '
+                              'with parsl. This flag will check the following '
```
Oddant1, line 230: This comment isn't finished because we discussed the exact locations and the order in which it checks them potentially being in flux. We also discussed putting the default config in a vendored file, but I'm not sure exactly where we would want that to be.

q2cli/commands.py
```diff
@@ -336,6 +348,21 @@ def __call__(self, **kwargs):
             raise ValueError('Cannot set a pool to be used for recycling and '
                              'no recycle simultaneously.')
 
+        parsl = kwargs.pop('parsl', False)
+        parsl_config_fp = kwargs.pop('parsl_config', None)
+
+        if parsl and parsl_config_fp is not None:
+            raise ValueError('Cannot use both --parsl and --parsl-config. '
```
ebolyen, line 355: Let's make it so that `--parsl-config` is an optional extra that overrides the default config.

q2cli/commands.py
```diff
@@ -402,6 +429,9 @@ def __call__(self, **kwargs):
         cleanup_logfile = False
         try:
             with qiime2.util.redirected_stdio(stdout=log, stderr=log):
+                if parsl:
+                    action = action.parsl
+
                 if recycle_pool is None:
                     results = action(**arguments)
```
ebolyen, line 436: Missing another `if parsl` to block on results

q2cli/commands.py
```diff
@@ -415,6 +445,13 @@ def __call__(self, **kwargs):
                     pool = cache.create_pool(key=recycle_pool, reuse=True)
                     with pool:
                         results = action(**arguments)
+
+                        # If we executed in a pool using parsl we need to get
+                        # our results inside of the context manager to ensure
+                        # that the pool is set for the entirety of the
+                        # execution
+                        if parsl:
+                            results = results.result()
```
ebolyen, line 454: If you change `result()` to `_result()`, don't forget to update

ebolyen (changes requested):

q2cli/commands.py
```diff
@@ -225,6 +225,30 @@ def __init__(self, name, plugin, action):
...
+                              'appdirs.user_config_dir(qiime2)/'
+                              'qiime2_config.toml.'
+                              '\n\n3. It will check for a file at '
+                              'appdirs.site_config_dir(qiime2)/'
+                              'qiime2_config.toml.'
+                              '\n\n4. It will check for a file at '
+                              '$CONDA_PREIFX/etc/qiime2_config.toml'
+                              '\n\n5. If none of the others are found, it '
+                              'will write the default config to $CONDA_PREFIX/'
+                              'etc/qiime2_config.toml and use that.'),
```
Oddant1, line 246: This help text flows in a really gross way when it actually prints in the terminal. Will look into fixing it.

ebolyen (approved):

PR #302, opened 2023-07-05
NEW: adds ResultCollection to Usage Driver

lizgehret:
CLI pairing to [#691](https://github.com/qiime2/qiime2/pull/691)

Co-Authored by: @Oddant1

### DIFF
--- a/q2cli/core/usage.py
+++ b/q2cli/core/usage.py
@@ -47,11 +47,17 @@ def write_plugin_example_data(plugin, output_dir):
 class CLIUsageVariable(usage.UsageVariable):
     EXT = {
         'artifact': '.qza',
-        'result_collection': '/',
+        'artifact_collection': '/',
+        'visualization_collection': '/',
         'visualization': '.qzv',
         'metadata': '.tsv',
         'column': '',
         'format': '',
+    }
+
+    ELEMENT_EXT = {
+        'artifact_collection': '.qza',
+        'visualization_collection': '.qzv'
     }
 
     @property
@@ -61,6 +67,13 @@ class CLIUsageVariable(usage.UsageVariable):
     @staticmethod
     def to_cli_name(val):
         return util.to_cli_name(val)
+
+    def _key_helper(self, input_path, key):
+        if self.var_type not in self.ELEMENT_EXT:
+            raise KeyboardInterrupt(
+                f'Cannot key non-collection type {self.var_type}')
+
+        return "%s%s%s" % (input_path, key, self.ELEMENT_EXT[self.var_type])
 
     def to_interface_name(self):
         if hasattr(self, '_q2cli_ref'):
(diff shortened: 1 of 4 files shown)


lizgehret:
Current test failure wasn't happening locally using the current dev env, so seems like the issue is from a differing dependency version in 2023.5. Currently investigating.

q2cli/core/usage.py
```diff
@@ -47,6 +47,10 @@ def write_plugin_example_data(plugin, output_dir):
 class CLIUsageVariable(usage.UsageVariable):
     EXT = {
         'artifact': '.qza',
+        # it would be nice to have a / as the 'ext' for collections so that
+        # it's clear that it's a directory, but that's proving to be a pain
+        # so putting on hold for now
```
ebolyen, line 52: Agreed, what seems to be the issue with that?

ebolyen (commented):
This all looks about right to me (still pending assertions from the other PR).

q2cli/core/usage.py
lizgehret: Screen Shot 2023-07-06 at 11 21 42 AM

@ebolyen this is what happens when adding '/' as the extension for result_collection, and including '/' after the input and output collection dirs within the usage tests. We weren't confident that there was a straightforward way to handle, but this would definitely be the ideal solution.

ebolyen: It's probably pretty easy, the checker lives on the Click Type, which would have enough context to know if it was a collection, and change this check as appropriate: https://github.com/qiime2/q2cli/blob/master/q2cli/click/type.py#L96-L98

ebolyen: I think it's actually the line above what I linked that over-interprets the trailing slash

ebolyen: FWIW, this is actually something a user would run into as well, because if they added a reasonable trailing slash, they would get the same error, so this is actually the usage API giving us a better range of tests.

q2cli/core/usage.py
```diff
@@ -74,14 +75,18 @@ def to_interface_name(self):
...
             return
 
         INDENT = self.use.INDENT
         input_path = self.to_interface_name()
         expr = shlex.quote(expression)
 
+        if key:
+            input_path = \
+                "%s%s%s" % (input_path, key, self.execute()[key].extension)
+
```
Oddant1, line 89: @lizgehret I realized something after you left. 

1. I lied and all elements in the collection might not be the same type. It will be for an output, but not necessarily for an input. 

2. We have the dict. We can get the item we want from the dict. We should probably just get that item and read its extension and not do the goofy thing we were doing before.

q2cli/core/usage.py
```diff
@@ -74,14 +75,18 @@ def to_interface_name(self):
...
         if not self.use.enable_assertions:
             return
 
         INDENT = self.use.INDENT
         input_path = self.to_interface_name()
         expr = shlex.quote(expression)
 
+        if key:
+            input_path = \
+                "%s%s%s" % (input_path, key, self.execute()[key].extension)
```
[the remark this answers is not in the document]
lizgehret: Yeah, it seemed like overkill to us too - but the Usage API has no way of knowing what the output type is until the factory has been executed. I think we'd have to add some new functionality that would allow for that, but not sure if the juice is worth the squeeze. We'll see what we can do this afternoon and circle back if there isn't a promising solution.

ebolyen (approved):
Looks great!

lizgehret (commented):
recent additions look good here too!

PR #303, opened 2023-07-12
Fixes here to go along with framework changes

Oddant1:
Changes to go along with qiime2/qiime2#695. Blocked by qiime2/qiime2#695

### DIFF
--- a/q2cli/tests/data/mapping_config.toml
+++ b/q2cli/tests/data/mapping_config.toml
@@ -4,15 +4,12 @@ strategy = "None"
 [[parsl.executors]]
 class = "ThreadPoolExecutor"
 label = "default"
-max_threads = 6
+max_threads = 1
 
 [[parsl.executors]]
-class = "HighThroughputExecutor"
-label = "htex"
-max_workers = 6
-
-[parsl.executors.provider]
-class = "LocalProvider"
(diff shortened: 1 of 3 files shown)


q2cli/tests/data/mapping_config.toml
```diff
@@ -4,15 +4,12 @@ strategy = "None"
 [[parsl.executors]]
 class = "ThreadPoolExecutor"
 label = "default"
-max_threads = 6
+max_threads = 1
```
lizgehret, line 7: Does this help to resolve the socket leaks?

lizgehret (approved):
looks reasonable to me

q2cli/tests/data/mapping_config.toml
Oddant1: No, but I had hoped it would when I was using HighThroughputExecutors.. ThreadPoolExecutors don't open ports, but HighThroughputExecutors do, the solution to the problem ended up being not using HighThroughputExecutors at all. This decrease is still useful though because there was no good reason to use a bunch of threads. It was just a waste of resources.

lizgehret (approved):
Tested locally, and once the framework PR is merged this is also good to go.

PR #306, opened 2023-08-28
Hide actions that start with _ and have flag to show them

### LINKED ISSUE #301: q2cli should hide actions with names that start with `_`
- `_` shouldn't be replaced with `-`
- actions starting with underscores should be hidden by default, with a flag to include them in the list when called as `qiime plugin-name`

Oddant1:
Closes #301. Hides actions that start with '_' by default and adds a flag to display them.

~~For click related reasons that I do not fully understand yet, this does not work and passing the command results in `Error: Missing Command`  think I more or less understand why it's happening, but I don't have a good solution for the issue yet.~~

Should be gtg now.

### DIFF
--- a/q2cli/commands.py
+++ b/q2cli/commands.py
@@ -158,7 +158,7 @@ class PluginCommand(BaseCommandMixin, click.MultiCommand):
                     help="This plugin has hidden actions with names starting "
                          "with '_'. These are generally called internally by "
                          "pipelines. Passing this flag will display those "
-                         "methods."))
+                         "actions."))
 
         super().__init__(name, *args, short_help=plugin['short_description'],
                          help=help_, params=params, **kwargs)
--- a/q2cli/commands.py
+++ b/q2cli/commands.py
@@ -221,6 +221,8 @@ class PluginCommand(BaseCommandMixin, click.MultiCommand):
 
     def get_command(self, ctx, name):
         try:
+            # Hidden actions are still valid commands
+            self._action_lookup.update(self._hidden_actions)
             action = self._action_lookup[name]
         except KeyError:
             from q2cli.util import get_close_matches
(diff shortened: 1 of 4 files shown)


Oddant1:
~~blocked by qiime2/qiime2#710~~

Oddant1:
~~This PR does not currently allow you to actually execute the hidden method from the command line. Need to fix that~~

q2cli/util.py
```diff
@@ -36,6 +36,15 @@ def get_completion_path():
     return os.path.join(get_cache_dir(), 'completion.sh')
 
 
+def hidden_to_cli_name(name):
+    # Safety first
+    assert name.startswith('_')
```
gregcaporaso: I'm assuming you always want this to raise an error if the name doesn't start with `_` - is that right? If so, it's better to use if/raise as [`assert` statements aren't always evaluated](https://stackoverflow.com/a/1838411/3424666).

q2cli/commands.py
```diff
@@ -171,6 +189,28 @@ def _get_citation_records(self):
...
+        # which for this is False. Honestly not 100% sure why we need that
+        # here, but it is in the check in _get_version, and I a mimicking that
+        if not value or ctx.resilient_parsing:
+            return
+
+        from click.utils import echo
+
+        self._action_lookup.update(self._hidden_actions)
+
+        # Handle the printing and exiting here. This feels like a pretty
```
gregcaporaso, line 208: I'm not familiar enough with click to know if there is a better way. Maybe worth getting some quick input from @ebolyen on this one.

gregcaporaso (commented):
In general this looks good to me - couple of comments.

q2cli/commands.py
ebolyen: Nope, this is exactly the right API it seems.

ebolyen (approved):
Looks great, thanks @Oddant1!

ebolyen:
Tests pass locally with the latest framework, so merging now.

PR #307, opened 2023-09-07
MAINT: Ensure we use the proper cache when no_recycle set

Oddant1:
Fixes a bug where using `--no-recycle` and `--use-cache` would cause the set cache to not be used,

### DIFF
--- a/q2cli/commands.py
+++ b/q2cli/commands.py
@@ -535,19 +535,20 @@ class ActionCommand(BaseCommandMixin, click.Command):
 
     def _execute_action(self, action, arguments, recycle_pool=None,
                         cache=None):
-        if recycle_pool is None:
-            results = action(**arguments)
-            results = results._result()
-        else:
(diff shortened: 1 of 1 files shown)


q2cli/commands.py
```diff
@@ -535,19 +535,20 @@ def __call__(self, **kwargs):
 
     def _execute_action(self, action, arguments, recycle_pool=None,
                         cache=None):
-        if recycle_pool is None:
-            results = action(**arguments)
-            results = results._result()
-        else:
-            pool = cache.create_pool(key=recycle_pool, reuse=True)
-            with pool:
+        with cache:
```
ebolyen, line 537: What if `cache` is `None`? Is a value always passed at this point?

Oddant1: Yes that does not need to be a kwarg we will always have a cache passed in either the one they set or the default cache

lizgehret (approved):

PR #309, opened 2023-09-20
NEW: Integrate provenance lib into cli

colinvwood:


colinvwood:
blocked by qiime2/qiime2#712

q2cli/builtin/tools.py
```diff
@@ -907,3 +908,249 @@ def cache_status(cache):
     success = "Status of the cache at the path '%s':\n\n%s" % \
         (cache, output)
     click.echo(CONFIG.cfg_style('success', success))
+
+
+@tools.command(name='replay-provenance', cls=ToolCommand)
+@click.option('--in-fp', required=True,
+              help='filepath to a QIIME 2 Archive or directory of Archives')
```
gregcaporaso:
```suggestion
              help='filepath to a QIIME 2 Archive (.qza/.qzv) or '
                   'directory of Archives')
```

Useful to define this here, in case the user isn't familiar with the term `Archive` in this context.

gregcaporaso: Not sure if the line break is needed or not for linting.

q2cli/builtin/tools.py
```diff
@@ -907,3 +908,249 @@ def cache_status(cache):
...
+@click.option('--recurse/--no-recurse',
+              default=False,
+              show_default=True,
+              help=('if in-fp is a directory, will also search sub-directories'
+                    ' when finding .qza/.qzv files to parse'))
```
gregcaporaso:
```suggestion
                    ' when finding Archives to parse'))
```

Since we define `Archive` in the previous help text now.

q2cli/builtin/tools.py
```diff
@@ -907,3 +908,249 @@ def cache_status(cache):
...
+        dump_recorded_metadata=dump_recorded_metadata,
+        md_out_fp=metadata_out_fp
+    )
+    filename = os.path.realpath(out_fp)
+    click.echo(f'{usage_driver} replay script written to {filename}')
+
+
+@tools.command(name='replay-citations', cls=ToolCommand)
+@click.option('--in-fp', required=True,
+              help='filepath to a QIIME 2 Archive or directory of Archives')
```
gregcaporaso: I recommend assigning the `help` text that you're re-using for options that show up in multiple commands to variables, and then providing those variables when defining options. That would let my suggestions above (and any future edits) propagate to all commands that use those options.

gregcaporaso (changes requested):
@colinvwood, I am primarily reviewing the new commands and the structure here (since I believe the code/tests are generally just moved over), and skimmed over all the rest. Let me know if that are aspects that you want a detailed review on other than that. This looks good to me - just a few minor suggestions.

q2cli/builtin/tools.py
```diff
@@ -907,3 +908,249 @@ def cache_status(cache):
...
+              help='check that replayed archives are intact and uncorrupted')
+@click.option('--parse-metadata/--no-parse-metadata',
+              default=True,
+              show_default=True,
+              help=('parse the original metadata captured in provenance '
+                    'for review or replay'))
+@click.option('--use-recorded-metadata/--no-use-recorded-metadata',
+              default=False,
+              show_default=True,
+              help='re-use the original metadata captured in provenance')
```
Oddant1: Why is the default behavior to parse the metadata but not use it?

q2cli/builtin/tools.py
```diff
@@ -907,3 +908,249 @@ def cache_status(cache):
     success = "Status of the cache at the path '%s':\n\n%s" % \
         (cache, output)
     click.echo(CONFIG.cfg_style('success', success))
+
+
+@tools.command(name='replay-provenance', cls=ToolCommand)
+@click.option('--in-fp', required=True,
```
Oddant1: We could make this a [click.Path](https://github.com/qiime2/q2cli/blob/6c91089e0fad52b3558586f93cdb8f01b0379134/q2cli/builtin/dev.py#L78) exists=True, file_okay=True, dir_okay=True, readable=True

q2cli/builtin/tools.py
```diff
@@ -907,3 +908,249 @@ def cache_status(cache):
...
+              default=True,
+              show_default=True,
+              help='print status messages to stdout while processing')
+@click.option('--dump-recorded-metadata/--no-dump-recorded-metadata',
+              default=True,
+              show_default=True,
+              help='write the original metadata captured in provenance to '
+                   'disk in the --metadata-out-fp directory')
+@click.option('--metadata-out-fp',
+              default='',
```
Oddant1, line 977: Seems like we could use one of these [`OutputDirType`](https://github.com/qiime2/q2cli/blob/6c91089e0fad52b3558586f93cdb8f01b0379134/q2cli/commands.py#L262) things for this. This would however enforce that the dir we give does not exist yet which this does not currently enforce. It says that's recommended, but I'm not sure if enforcing that is desirable (and then obviously the docs would need updated).

q2cli/builtin/tools.py
```diff
@@ -907,3 +908,249 @@ def cache_status(cache):
...
+              show_default=True,
+              help=('the directory where captured study metadata '
+                    'should be written if --dump-recorded-metadata. This '
+                    'often produces many outputs, so a dedicated directory '
+                    'should generally be used. Creates the directory if it '
+                    'does not already exist. By default, metadata is written '
+                    'to `${PWD}/recorded_metadata/`'))
+@click.option('--out-fp',
```
Oddant1, line 985: We could make this a [click.Path](https://github.com/qiime2/q2cli/blob/6c91089e0fad52b3558586f93cdb8f01b0379134/q2cli/builtin/dev.py#L78) exists=False, writable=True

q2cli/builtin/tools.py
```diff
@@ -907,3 +908,249 @@ def cache_status(cache):
...
+                    'but introduces a small risk of reference loss.'))
+@click.option('--suppress-header/--no-suppress-header',
+              default=False,
+              show_default=True,
+              help='do not write header/footer blocks in the output file')
+@click.option('--verbose/--no-verbose',
+              default=True,
+              show_default=True,
+              help='print status messages to stdout while processing')
+@click.option('--out-fp',
```
Oddant1, line 1056: Same as other `--out-fp`, we could make this a [click.Path](https://github.com/qiime2/q2cli/blob/6c91089e0fad52b3558586f93cdb8f01b0379134/q2cli/builtin/dev.py#L78) exists=False, writable=True

q2cli/builtin/tools.py
```diff
@@ -907,3 +908,249 @@ def cache_status(cache):
...
+@click.option('--verbose/--no-verbose',
+              default=True,
+              show_default=True,
+              help='print status messages to stdout while processing')
+@click.option('--dump-recorded-metadata/--no-dump-recorded-metadata',
+              default=True,
+              show_default=True,
+              help='write the original metadata captured in provenance to '
+                   'disk in the --out-fp archive')
+@click.option('--out-fp',
```
Oddant1, line 1130: Same as other `--out-fp`, we could make this a [click.Path](https://github.com/qiime2/q2cli/blob/6c91089e0fad52b3558586f93cdb8f01b0379134/q2cli/builtin/dev.py#L78) exists=False, writable=True

q2cli/builtin/tools.py
```diff
@@ -907,3 +908,249 @@ def cache_status(cache):
...
+        deduplicate=deduplicate,
+        suppress_header=suppress_header
+    )
+    filename = os.path.realpath(out_fp)
+    click.echo(f'citations bibtex file written to {filename}')
+
+
+@tools.command(name='replay-supplement', cls=ToolCommand)
+@click.option('--in-fp', required=True,
+              help='filepath to a QIIME 2 Archive or directory of Archives')
```
Oddant1, line 1092: Same as other `--in-fp`, we could make this a [click.Path](https://github.com/qiime2/q2cli/blob/6c91089e0fad52b3558586f93cdb8f01b0379134/q2cli/builtin/dev.py#L78) exists=True, file_okay=True, dir_okay=True, readable=Tru

Oddant1:
@colinvwood actually now that I think about it again do not change the `--in-fp` options above. I think we probably would like to be able to hand in cached artifacts and we don't want click validating that our cache:key is a valid fp because it won't be (it will still take more work to make prov-replay work on cached artifacts though)

colinvwood:
> @colinvwood actually now that I think about it again do not change the `--in-fp` options above. I think we probably would like to be able to hand in cached artifacts and we don't want click validating that our cache:key is a valid fp because it won't be (it will still take more work to make prov-replay work on cached artifacts though)

you mean the `--out-fp`s?

Oddant1:
@colinvwood no you should be good to do the out-fps I have comments on the in-fps that I marked resolved so if you aren't seeing them that's why

q2cli/builtin/tools.py
```diff
@@ -907,3 +908,274 @@ def cache_status(cache):
...
+    from qiime2.core.archive.provenance_lib.replay import replay_provenance
+    from qiime2.sdk.util import get_available_usage_drivers
+
+    usage_drivers = get_available_usage_drivers()
+    try:
+        usage_driver_type = usage_drivers[usage_driver]
+    except KeyError:
+        msg = (
+            f'The {usage_driver} usage driver is not available in the '
+            'current evnironment.'
```
lizgehret:
```suggestion
            'current environment.'
```

lizgehret (approved):
nice work! testing locally but generally lgtm.

q2cli/builtin/tools.py
```diff
@@ -907,3 +908,274 @@ def cache_status(cache):
...
+              help=replay_suppress_header_help)
+@click.option('--verbose/--no-verbose',
+              default=True,
+              show_default=True,
+              help=replay_verbose_help)
+@click.option('--dump-recorded-metadata/--no-dump-recorded-metadata',
+              default=True,
+              show_default=True,
+              help=replay_dump_recorded_metadata_help)
+@click.option('--metadata-out-fp',
```
lizgehret: This parameter might be more accurately named as `--metadata-out-dir` so that it's clear that this output is written to a directory path (while `--out-fp` is just written to a file path).

lizgehret (commented):
@colinvwood and I discussed modifying the help text for `dump-recorded-metadata` in replay and replay-supplement. Other than that, local testing all works great!

lizgehret (approved):
lgtm, those descriptions are much clearer - thanks @colinvwood!

PR #318, opened 2023-10-06
Add qiime tools cache-import

### LINKED ISSUE #317: enable direct-to-cache imports
To speed up import of very large data sets, we should be able to directly import data (e.g., a collection of `fastq.gz` files) to an artifact cache, without going through an Archive (`.qza`). This would avoid the zipping of the large dataset.

Oddant1:
Closes #317 by adding the `qiime tools cache-import` command


Oddant1:
For reasons I do not fully understand this appears to be ~~blocked by #319~~

lizgehret (approved):
lgtm

ebolyen:
This also LGTM, and this new CLI tool would be a good place to put solutions to:
https://github.com/qiime2/qiime2/issues/730
and
https://github.com/qiime2/qiime2/issues/721

gregcaporaso:
Ok, I'm going to merge in the next few minutes since we have approvals from @lizgehret and @ebolyen, unless I hear otherwise. Thanks for the code and the reviews!

PR #319, opened 2023-10-09
Make things aware of other-plugin

Oddant1:
Make the cli aware of the new "other-plugin" that @colinvwood added


colinvwood (approved):

colinvwood:
Note: as stated in #318, these changes seem to be necessary for CI to pass. Regardless however, these changes should be added because they mirror those made in qiime2/qiime2#712.

PR #322, opened 2023-10-18
Expand ~

### LINKED ISSUE #308: De-Facto keyed collections do not like paths starting with ~
**Bug Description**
When creating a de-facto keyed collection of inputs on the command-line paths that start with `~` do not work. It does not find the artifact then errors out on the key as if it were a bad cache. We need to expand the `~` to be the home directory if it is at the head of the path. Click does this on its own if it just receives a path like `~/foo`, but if it gets a key:path like `bar:~/foo` it obviously doesn't understand that it needs to expand that `~`.

1. Pick an action that takes a Collection of inputs
2. Give the collection a keyed member with a path starting with `~`
3. See it get confused

**Expected behavior**
It should actually find the artifact.

**Screenshots**
You can see that it found the `foo` keyed artifact, so the artifact does exist, but for the `bar` keyed one the path starting with `~` confused it into not finding the artifact.
___
image
___

**Computation Environment**
- OS: Linux Mint 21.2
- QIIME 2 Latest Dev as of time of writing

Oddant1:
Closes #308 by expanding the `~` using `os.path.expanduser` allowing keyed de-facto collection elements to use a path starting with `~`.


gregcaporaso (changes requested):
@Oddant1, any reason not to include a new test for this case? Aside from that this looks good.

Oddant1:
@gregcaporaso we would need to write to the user's home directory which I wanted to avoid doing. There may be some sort of work around for that, I didn't look too hard for one.

gregcaporaso:
That makes sense. How about confirming that the resulting path starts with `os.environ['HOME']`? (And I realize it's incredibly unlikely that this single line of code you're adding is buggy - I'm just aiming to prevent against a regression where someone removes this line because they assume Click is going to handle it.)

Oddant1:
@gregcaporaso I just added a kinda real test, seemed like the thing to do. It wasn't difficult to work around writing to home which I figured it wouldn't be. Probably should have just done it already.

q2cli/tests/test_cli.py
```diff
@@ -926,5 +948,43 @@ def test_empty_directory(self):
...
+    stomped = (set(update.keys()) | set(remove)) & set(env.keys())
+    # Environment variables and values to restore on exit.
+    update_after = {k: env[k] for k in stomped}
+    # Environment variables and values to remove on exit.
+    remove_after = frozenset(k for k in update if k not in env)
+
+    try:
+        env.update(update)
+        [env.pop(k, None) for k in remove]
+        yield
```
gregcaporaso, line 983: The try/yield/finally is a cool pattern - I don't think I've seen that before.

gregcaporaso (approved):
Looks good, thank you!

PR #323, opened 2023-11-08
add usage driver support for replaying result collections

colinvwood:


colinvwood:
~~blocked by qiime2/qiime2#729~~

q2cli/tests/test_usage.py
```diff
@@ -197,3 +198,34 @@ def test_round_trip(action, example):
...
+
+
+class ReplayResultCollectionTests(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.plugin = get_dummy_plugin()
+
+    @classmethod
+    def tearDownClass(cls):
+        pass
```
colinvwood: 🤨

q2cli/core/usage.py
```diff
@@ -173,6 +173,80 @@ def init_artifact_collection(self, name, factory):
...
+        for key in keys:
+            keys_arg += f'{key} '
+        keys_arg += ')'
+        names_arg = '( '
+        for name in names:
+            names_arg += f'{name} '
+        names_arg += ')'
+
+        lines = [
+            '## ignore: constructing result collection ##',
```
ebolyen: Is this really something to ignore for the reader?

q2cli/tests/test_usage.py
```diff
@@ -197,3 +198,30 @@ def test_round_trip(action, example):
...
+construct_result_collection
+##
+qiime dummy-plugin dict-of-ints \\
+  --i-ints rc-in/ \\
+  --o-output rc-out/
+## ignore: accessing result collection member ##
+ln -s rc-out/b.qza ints-b-from-collection.qza
+##"""
+
+        self.assertEqual(exp, use.render())
```
ebolyen, line 227: Shouldn't the header show up from the render call?

q2cli/core/usage.py
```diff
@@ -173,6 +173,80 @@ def init_artifact_collection(self, name, factory):
 
         return variable
 
+    def construct_collection(self, name, member_type, members):
+        # make sure members is dict to avoid repeated type checking
+        if type(members) is list:
+            members = {str(i): member for i, member in enumerate(members)}
```
ebolyen: Yeah this definitely belongs in the parent class at this point I think.

q2cli/core/usage.py
```diff
@@ -173,6 +173,80 @@ def init_artifact_collection(self, name, factory):
...
+
+        variable = super().construct_collection(name, member_type, members)
+
+        str_namespace = {str(name) for name in self.namespace}
```
ebolyen: `self.namespace` is already a set of strings.

q2cli/core/usage.py
```diff
@@ -173,6 +173,80 @@ def init_artifact_collection(self, name, factory):
...
+            msg = (
+                f'{diff} not found in driver\'s namespace. Make sure '
+                'that all ResultCollection members have been properly '
+                'created.'
+            )
+            raise ValueError(msg)
+
+        # make rc dir
+        rc_dir = variable.to_interface_name()
+        if os.path.exists(rc_dir):
```
ebolyen: The usage example render isn't necessarily running on the host or directory that will be running the script. So this check isn't as useful as you might hope.

q2cli/core/usage.py
```diff
@@ -173,6 +173,80 @@ def init_artifact_collection(self, name, factory):
...
+        )
+
+        rc_dir = collection.to_interface_name()
+        if collection.var_type == 'artifact_collection':
+            member_fp = os.path.join(rc_dir, f'{key}.qza')
+        else:
+            member_fp = os.path.join(rc_dir, f'{key}.qzv')
+
+        lines = [
+            '## ignore: accessing result collection member ##',
```
ebolyen: I don't think this is something to ignore for the reader either.

ebolyen (changes requested):

q2cli/tests/test_usage.py
colinvwood: `build_header` and `build_footer` are methods that aren't called here

q2cli/core/usage.py
```diff
@@ -173,6 +173,67 @@ def init_artifact_collection(self, name, factory):
...
+        diff = set(
+            member.to_interface_name() for member in members.values()
+        ) - str_namespace
+        if diff:
+            msg = (
+                f'{diff} not found in driver\'s namespace. Make sure '
+                'that all ResultCollection members have been properly '
+                'created.'
+            )
+            raise ValueError(msg)
```
ebolyen: This is in the base-class now right? So we can delete this chunk.

q2cli/tests/test_usage.py
ebolyen: Got it, that's part of the replay API, this will be something to keep in mind if collections get constructed in other non-replay usage examples. But we can revisit that in the future.

ebolyen (changes requested):
Basically good to go.

ebolyen (approved):

PR #328, opened 2023-12-20
TST: fix roundtrip on /bin/sh

ebolyen:
I believe we're running into an issue with the format of the subprocess run when testing with the new collection stuff:
https://github.com/qiime2/distributions/actions/runs/7281173631/job/19841955368?pr=62#step:8:267.

This change will _probably_ fix the issue, as shell=True is usually intended to be done with a single string. It hasn't come up as an issue but I think the new collection syntax is finally making it break (in /bin/sh only, bash works fine, hence not seeing it locally).

### DIFF
--- a/q2cli/tests/test_usage.py
+++ b/q2cli/tests/test_usage.py
@@ -193,7 +193,7 @@ def test_round_trip(action, example):
     with tempfile.TemporaryDirectory() as tmpdir:
         for ref, data in use.get_example_data():
             data.save(os.path.join(tmpdir, ref))
-        subprocess.run([rendered],
+        subprocess.run(rendered,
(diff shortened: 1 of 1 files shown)


ebolyen:
actually, I don't think this is it either

ebolyen:
Seems like it's going to depend on your `sh` implementation, but often arrays aren't supported as they didn't exist in the original.

https://unix.stackexchange.com/questions/253892/syntax-error-unexpected-when-creating-an-array

So I guess we need to detect if it's linux and use `/bin/bash` as the shell executable for `subprocess.run` as the behavior on GH workers is just not suitable. (I would like the keep zsh around on osx, as that's useful.)

PR #329, opened 2024-01-10
BUG: missing metadata column error not raised correctly

### LINKED ISSUE #327: BUG: CLI says missing metdata-file when a metadata column is not passed in.
**Improvement Description**
When there is a required metadata column and you only pass in the metadata-file, you get an error that says "(1/1) Missing option '--m-metadata-file'." . This feels like a confusing error because I am actually missing a metadata column. 

**Recreation Steps** 
1. Run Beta group significance or another method that requires a  --m-metadata-column. 
2. Input a metadata file but not a metadata column
3. see error 😱 

**Current Behavior**

When a metadata column is required but not passed in the error says you are missing the metadata file.
Screenshot 2023-12-12 at 8 19 12 AM

**Proposed Behavior**
QIIME should tell users that they are missing a metadata column instead of telling users they are missing a metadata file. 

**Comments**
1. I believe this is a cli issue because in the python interface has you pass in a metadata column so this error will only occur in the cli.
2. I have found this error in `diversity beta-group-significance` and `sample-classifier fit-classifier`

lizgehret:
Fixes #327

### DIFF
--- a/q2cli/tests/test_cli.py
+++ b/q2cli/tests/test_cli.py
@@ -569,7 +569,7 @@ class TestMetadataSupport(MetadataTestsBase):
 
 
 class TestMetadataColumnSupport(MetadataTestsBase):
-    def test_required_missing(self):
+    def test_required_missing_file(self):
         result = self._run_command(
             'identity-with-metadata-column', '--i-ints', self.input_artifact,
             '--o-out', self.output_artifact)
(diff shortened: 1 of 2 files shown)


q2cli/tests/test_cli.py
```diff
@@ -569,7 +569,7 @@ def test_invalid_metadata_merge(self):
 
 
 class TestMetadataColumnSupport(MetadataTestsBase):
-    def test_required_missing(self):
+    def test_required_missing_file(self):
```
Oddant1: We have this test here where we pass neither thing and now we have a test where we pass in the file but not the column, do we want one where we pass in the column but not the file? Does that even serve a purpose? It doesn't feel like it should honestly but given the behavior we've seen here. . .

lizgehret: yeah that's fair enough @Oddant1, can't hurt to add a test for that and maybe it will help us catch something dumb that we aren't expecting in the future. i'll get that added!

colinvwood (approved):

Oddant1 (approved):

lizgehret:
Alright I'm going to merge this since I have approvals from @Oddant1 and @colinvwood. Thanks for the reviews y'all!

PR #337, opened 2024-05-28
BUG: Update tests to use ReplayNamespaces not NamespaceCollection

Oddant1:
Ran into this failure locally induced by merging qiime2/qiime2#739 resolved it


lizgehret (approved):

lizgehret:
this is failing due to a stale qiime2 build in the conda build config - @Oddant1 and I both tested locally and all tests pass, so this isn't an interesting (or real) failure. going to merge this and then we should have a passing prepare

PR #351, opened 2024-12-12
Vendored view cli

Oddant1:
Cli side of qiime2/q2view#25


q2cli/builtin/tools.py
```diff
@@ -511,72 +511,141 @@ def _merge_metadata(paths):
...
+
+                    with open(file_path, 'rb') as file:
+                        self.wfile.write(file.read())
+            # Otherwise default to super class. This will respond appropriately
+            # to requests for assets that are part of the vendored view app and
+            # will reject any requests for files outside the served directory
+            else:
+                super().do_GET()
+
+    VENDOR_PATH = '/home/anthony/src/qiime2/q2view/vendored/'
```
ebolyen: What we are going to need to do is add q2view to the build process for q2cli. We can handle that in the Makefile (this will be a bit similar to how we build the vendored visualizations in q2-diversity, so that is a pretty ok example).

q2cli/builtin/tools.py
```diff
@@ -511,72 +511,141 @@ def _merge_metadata(paths):
...
                cls=ToolCommand)
-@click.argument('visualization-path', metavar='VISUALIZATION',
+@click.argument('result-path', metavar='RESULT',
                 type=click.Path(file_okay=True, dir_okay=False, readable=True))
-@click.option('--index-extension', required=False, default='html',
-              help='The extension of the index file that should be opened. '
-                   '[default: html]')
-def view(visualization_path, index_extension):
+@click.option('--port', required=False, type=click.IntRange(1024, 65535),
+              default=get_free_port(),
```
ebolyen: I think the default should be `None` here, and then when None, you call the above `get_free_port()`, there will also be a race condition technically, as I think the port is released as soon as you close the socket, but I'm not sure if there's any way around that (i.e. can you recycle a socket with HTTPHandler? probably not).

Oddant1: Good point. I didn't really consider that.

q2cli/builtin/tools.py
Oddant1: @ebolyen how do you feel about making q2view a submodule of q2cli to accomplish this?

ebolyen:
Ok, I like the submodule approach I think.

**Acceptance Criteria**:
- [x] `make dev` will build the vendored q2view and write it to an in-package `q2cli/assets/` (or `q2cli/vendor/`) directory which `q2cli` will then use once installed.
- [x] `make install` will do the same steps to vendor q2view as `make dev`. It should be verified that a conda build of the recipe includes the working vendored copy.
- [x] The source code of `q2view` should not be in the installed package itself, only the built JS application.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "vendored_q2view"]
+	path = vendored_q2view
```
Oddant1, line 2: This path is the branch this is pointing at and will need to be changed when qiime2/q2view#25 is merged

ebolyen:
Just played with this locally, pretty nice!

I think it would be helpful to silence some of the stdout and to print the URL that the user should visit to initiate the session. That way if the open handler doesn't work, the user can still open that URL (perhaps in a different browser than their default also).

Oddant1:
> Just played with this locally, pretty nice!
> 
> I think it would be helpful to silence some of the stdout and to print the URL that the user should visit to initiate the session. That way if the open handler doesn't work, the user can still open that URL (perhaps in a different browser than their default also).

Done

gregcaporaso:
Thanks for all of the work on this @Oddant1! I did some additional testing locally (since we last discussed) and all is working as expected.

PR #357, opened 2025-01-14
Try to get default pool test to fail less frequently

Oddant1:


Oddant1:
Well it will fail less frequently now...

PR #362, opened 2025-04-18
NEW: supporting commands for annotation actions on v7 archives

lizgehret:
blocked by https://github.com/qiime2/qiime2/pull/821

### DIFF
--- a/q2cli/builtin/tools.py
+++ b/q2cli/builtin/tools.py
@@ -1394,9 +1394,7 @@ def annotation_create(input_path, annotation_type,
     import qiime2.sdk
     from qiime2.core.annotate import Note
     from q2cli.core.config import CONFIG
-    from q2cli.util import get_plugin_manager
 
-    get_plugin_manager()
     # enforce exactly one of `--text` or `--file`
     if (text is None) == (filepath is None):
         raise click.UsageError('Exactly one of `--text` or `--file`'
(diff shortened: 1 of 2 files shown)


q2cli/builtin/tools.py
```diff
@@ -1237,3 +1237,138 @@ def supplement_replay(
...
+)
+def annotation_create(input_path, annotation_type,
+                      name, text, filepath, output_path):
+    """Load a Result, add an Annotation and save."""
+    import qiime2.sdk
+    from qiime2.core.annotate import Note
+    from q2cli.core.config import CONFIG
+    from q2cli.util import get_plugin_manager
+
+    get_plugin_manager()
```
Oddant1: I don't believe this is necessary

q2cli/builtin/tools.py
```diff
@@ -1237,3 +1237,138 @@ def supplement_replay(
...
+    help='Where to write the updated Result.'
+         '[default: overwrite input filepath]'
+)
+def annotation_remove(input_path, name, output_path):
+    """Load a Result, remove an Annotation by name and save."""
+    import qiime2.sdk
+    from q2cli.core.config import CONFIG
+    from q2cli.util import get_plugin_manager
+
+    get_plugin_manager()
```
Oddant1: I don't think this is necessary here either. We don't care about the type or format or anything of the result we are annotating

lizgehret: ok you're totally right - i gaslit myself on this bc i still had QIIMETEST set which is why none of the formats existed lol.

ebolyen:
drive by comment: it might be nice to have the input result as a click.argument, then you don't have to type `--input-path` (similar to qiime tools view) and there's no overlap with the note's filepath (should one be used)

lizgehret:
> drive by comment: it might be nice to have the input result as a click.argument, then you don't have to type `--input-path` (similar to qiime tools view) and there's no overlap with the note's filepath (should one be used)

@ebolyen has rescinded his comment due to the fact that outputdir is optional and exists. #myoutielovesoutputdirs

PR #363, opened 2025-04-20
minor help text updates

gregcaporaso:


q2cli/builtin/info.py
```diff
@@ -81,12 +81,13 @@ def info(config_level):
...
-    click.secho('To get help with QIIME 2, visit https://qiime2.org')
+    click.secho('To find help and learning resources, visit '
+                'https://qiime2.org.')
 
     if config_level:
         click.secho('To get help with configuring and/or understanding '
                     'QIIME 2 parallelization, visit '
                     'https://use.qiime2.org/en/latest/references/'
-                    'parallel-configuration.html')
+                    'parallel-configuration.html.')
```
ebolyen: The trailing period can break terminal links depending on the parser (just like in an email). It's probably best to leave it off.
```suggestion
                    'parallel-configuration.html')
```

PR #364, opened 2025-04-22
Be a good boy and end my headers so Chromium stops screaming at me

Oddant1:
Chromium doesn't like it when I don't end my headers


ebolyen:
Going to wait on this, because we'll need an update to the submodule checkout once https://github.com/qiime2/q2view/pull/37 is merged.

PR #370, opened 2025-05-29
MAINT: click stdout/stderr behavior and set pin

ebolyen:
Waiting on https://github.com/qiime2/distributions/pull/751

Fixes a future issue we will see where the `tab-qiime` script will run `qiime` to initialize things, however this has changed to print to stderr, which conda will interpret as a hook failure, boggling up the whole deal.


lizgehret (approved):
lgtm, i'm going to merge this and re-run tiny without a click pin to make sure this works as expected

PR #373, opened 2025-08-13
IMP: adds `cache-export` action

### LINKED ISSUE #350: exporting of Artifacts directly from an Artifact Cache does not work

Oddant1:
closes #350

### DIFF
--- a/q2cli/builtin/tools.py
+++ b/q2cli/builtin/tools.py
@@ -17,6 +17,24 @@ from q2cli.click.command import ToolCommand, ToolGroupCommand
 _COMBO_METAVAR = 'ARTIFACT/VISUALIZATION'
 
 
+def _import(type, input_path, input_format, validate_level):
+    import qiime2.sdk
+    import qiime2.plugin
+
+    try:
+        artifact = qiime2.sdk.Artifact.import_data(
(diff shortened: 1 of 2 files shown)


lizgehret (approved):
code review lgtm, going to do some user testing and then this should be good to merge!

lizgehret:
Okay @Oddant1 user testing generally looks good, I just have one syntax-related comment (and it's pretty minor so feel free to take it or leave it). It might be helpful to include a usage example in here for exporting so that it's clear to folks that you can't use the cache:key syntax to save the output to your cache. Even though this makes sense (bc the output won't be an artifact) I just wonder if that will confuse anyone who's used to that syntax when running other commands and saving output to cache in the cache:key format. Anyways, like I said it's pretty minor so happy to merge if you don't think the juice is worth the squeeze.

Oddant1:
@lizgehret We use `--cache` and `--key` as two separate params for all of the cache builtins, so I think it ought to be fine. I never did really think about whether it would be confusing for users or not though to have two different syntaxes... To me `--cache` and `--key` felt much more clear (and requires less futzing around on our end) `cache:key` only happened so we wouldn't have to massively disrupt the signatures of actions to use cached artifacts

PR #375, opened 2025-08-14
 LANG: 2025.10 updates

Oddant1:
Just make sure the parsl config is there at all instead of checking specifics because keeping up with parsl changes here would be annoying and pointless. We just care that it's there at all not the specific values (watch even the tiny bit we're checking here change with the next parsl release we use).


lizgehret (approved):
lgtm, merging now since this is currently compatible & will reduce re-testing post-2025.10 release

PR #376, opened 2025-09-04
TEST: add full test suite for annotations from archive 7.0

lizgehret:


q2cli/tests/test_tools.py
```diff
@@ -1277,43 +1278,379 @@ def test_replay_supplement_zipfile(self):
...
+        # annotation path for note from file
+        self.note_file = os.path.join(
+                self.tempdir, 'note_file.txt')
+        with open(self.note_file, 'w') as f:
+            f.write('my special text')
+
+    def tearDown(self):
+        shutil.rmtree(self.tempdir)
+
+    def test_annotation_commands_roundtrip_success(self):
```
Oddant1: I would probably have this assert things look how we expect on the artifact after every step, but I wouldn't consider that mandatory given the other tests.

Oddant1 (changes requested):
One comment and one actual change I think. Most of the annotation commands don't have the ToolCommand class which leads to the following when invoking them on the cli

image

The top one does have the ToolCommand class and looks correct. The others do not.

q2cli/tests/test_tools.py
lizgehret: we discussed this offline but i'm just going to remove this test in favor of the other more comprehensive tests (this was mainly just a stop gap test prior to adding the rest)

lizgehret:
> One comment and one actual change I think. Most of the annotation commands don't have the ToolCommand class which leads to the following when invoking them on the cli
> image
> 
> The top one does have the ToolCommand class and looks correct. The others do not.

oh thanks @Oddant1 that was definitely an oversight on my end. this has been fixed and should be ready for review round 2!

Oddant1:
lgtm

Oddant1 (approved):

PR #378, opened 2025-09-15
NEW: qiime tools make-report

ebolyen:
Goes with all the other report stuff.

- [x] Needs a test

### DIFF
--- a/q2cli/builtin/tools.py
+++ b/q2cli/builtin/tools.py
@@ -511,6 +511,38 @@ def _merge_metadata(paths):
     return metadata
 
 
+@tools.command(short_help='Make a QIIME 2 Report.',
+               help='Make a QIIME 2 report consisting of multiple '
+                    'visualizations. Visualizations must have unique '
+                    'filenames.',
+               cls=ToolCommand)
+@click.argument('visualization_paths', metavar='VIZ', required=True, nargs=-1,
+                type=click.Path(file_okay=True, dir_okay=False, readable=True))
+@click.option('--report-path', required=True, metavar='REPORT',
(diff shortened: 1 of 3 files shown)


ebolyen:
Depends on:
https://github.com/qiime2/qiime2/pull/852
https://github.com/qiime2/q2templates/pull/62

q2cli/builtin/tools.py
```diff
@@ -511,6 +511,38 @@ def _merge_metadata(paths):
...
+        click.echo(CONFIG.cfg_style('error', error), err=True)
+        click.get_current_context().exit(1)
+
+    entries = dict(zip(keys, loaded))
+
+    report = Visualization.make_report(matryoshka_template, entries)
+    result = report.save(report_path)
+    click.echo(CONFIG.cfg_style('success', f'Report saved to {result}'))
+
+
```
Oddant1, line 545: This will not accept cached visualizations or `Collection[Visualization]` output directories. Both of these do have easy workarounds that are just kinda inconvenient.

Oddant1: Also, I'm assuming the idea here is that if we create more templates, we will add a param here for the template to use, which will default to matryoshka? If so, maybe we should add a TODO or a note about that.

ebolyen: re: templates, probably not, but maybe? The CLI doesn't scale very well for configuring most of the reports you might imagine, but I could see something like a "linear-layout" in which case, yeah a flag would work.

re: cache, I figure that can be a future improvement when requested. Visualizations aren't usually too big and I didn't want to think about the filename logic for a cache key. It's not hard, but the CLI is getting the short end of the stick for this API anyways.

PR #379, opened 2025-09-30
Add Capture to dict

Oddant1:
Blocking qiime2/qiime2#873


ebolyen:
I think ideally we wouldn't even need this PR (or equivalent PRs for the other interfaces)

PR #381, opened 2025-10-08
NEW: archive 7.1

lizgehret:
adds support for Annotation[Signature]

### DIFF
--- a/q2cli/builtin/tools.py
+++ b/q2cli/builtin/tools.py
@@ -1458,7 +1458,7 @@ def supplement_replay(
 @click.option(
     '--annotation-type',
     required=True,
-    type=click.Choice(ANNOTATION_TYPE_DICT, case_sensitive=True),
+    type=click.Choice(list(ANNOTATION_TYPE_DICT.keys()), case_sensitive=True),
     help='Annotation type to create.'
 )
 @click.option(
--- a/q2cli/builtin/tools.py
+++ b/q2cli/builtin/tools.py
@@ -11,7 +11,7 @@ from typing import Literal
 
 import click
 
-from qiime2.core.annotate import ANNOTATION_TYPE_LIST
+from qiime2.core.annotate import ANNOTATION_TYPE_DICT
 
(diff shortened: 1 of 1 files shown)


q2cli/builtin/tools.py
```diff
@@ -1456,14 +1458,15 @@ def supplement_replay(
 @click.option(
     '--annotation-type',
     required=True,
-    type=click.Choice(['Note'], case_sensitive=True),
+    type=click.Choice(ANNOTATION_TYPE_DICT, case_sensitive=True),
```
Oddant1:
```suggestion
    type=click.Choice(list(ANNOTATION_TYPE_DICT.keys()), case_sensitive=True),
```

Does using just the entire dict here work as expected? I feel like there's likely to be some amount of unexpected behavior from this. I think what we want here is explicitly just a list of the keys from the dict, and we should probably explicitly get that.

q2cli/builtin/tools.py
```diff
@@ -1479,30 +1482,53 @@ def supplement_replay(
...
          '[default: overwrite input file]'
 )
-def annotation_create(input_path, annotation_type,
-                      name, text, filepath, output_path):
+def annotation_create(input_path, annotation_type, name,
+                      # note inputs
+                      text, filepath,
+                      # signature input
+                      fingerprint,
+                      output_path):
```
Oddant1, line 1503: Do we want a tool specifically called `signature_create`? @gregcaporaso do you have thoughts on that? Should we have `annotation_create` be used for all annotations, or should some types be spun off into their own tool? Particularly for something like signing, it feels like maybe that should get its own specific `signature_create` or `signature_add` tool.

lizgehret: I discussed this with @ebolyen a while back and we discussed it being fine to leave this as an annotation-wide action that can encapsulate all new annotation types, and then just keeping signature verification separate. I am open to segmenting this in the future if it ends up making more sense though.

PR #382, opened 2025-10-15
IMP: add migration warning on `ActionCommand` class

lizgehret:
~~blocked by https://github.com/qiime2/qiime2/pull/876~~

PR for user testing https://github.com/qiime2/q2-vizard/pull/83

### DIFF
--- a/q2cli/commands.py
+++ b/q2cli/commands.py
@@ -321,23 +321,38 @@ class ActionCommand(BaseCommandMixin, click.Command):
                 'warning', 'WARNING:\n\nThis command is deprecated and will '
                            'be removed in a future version of this plugin.'))
         elif self.action['migrated']:
-            mig = self.action['migrated']
-            if isinstance(mig, dict):
-                mig_help = (
-                    'WARNING:\n\nThis command will be migrated from '
-                    f'{q2cli.util.to_cli_name(self.plugin["name"])} in '
-                    f'{mig["from_distro"]} to {mig["to_plugin"]} in '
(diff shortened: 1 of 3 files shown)


q2cli/commands.py
```diff
@@ -315,10 +315,27 @@ def __init__(self, name, plugin, action):
...
                            'be removed in a future version of this plugin.'))
+        elif self.action['migrated']:
+            mig = self.action['migrated']
+            if isinstance(mig, dict):
+                mig_help = (
+                    'WARNING:\n\nThis command will be migrated from '
+                    f'{q2cli.util.to_cli_name(self.plugin["name"])} in '
+                    f'{mig["from_distro"]} to {mig["to_plugin"]} in '
+                    f'{mig["to_distro"]} in {mig["epoch"]}.'
+                )
```
Oddant1: This will have to change in response to the changes in the actual underlying API and what's on one of these migration dicts, like making some of the fields optional.

q2cli/commands.py
```diff
@@ -315,10 +315,27 @@ def __init__(self, name, plugin, action):
...
+            else:
+                mig_help = (
+                    'WARNING:\n\nThis command is slated for migration to '
+                    'another plugin/distribution in a future release.'
+                )
```
Oddant1: In practice, this will never be used because the framework validates the value that 'migrated' is set to, and if it is set to something illegal, the framework raises an exception.

Oddant1:
Passes for me locally (for future reference, CI failures are due to a framework PR that is merged but not in the CI env yet) and lgtm

PR #384, opened 2025-12-02
MAINT: CLI end of prov lib refactor

Oddant1:
Refactor prov-lib to use existing Archiver API and to be able to generate replay scripts without all actions in prov being in the environment.

Blocked by qiime2/qiime2#884


q2cli/core/usage.py
```diff
@@ -507,10 +508,14 @@ def _append_action_line(self, signature, param_name: str, value):
...
                 'Please see the docs and correct the parameter name '
                 'before running.\n')
-            cli_name = re.sub('_', '-', param_name)
-            line += self.INDENT + '--?-' + cli_name + ' ' + str(value)
-            line += ' \\'
-            self.recorder.append(line)
+            self._append_unknown_param(param_name, value)
+
+    def _append_unknown_param(self, param_name, value):
+        line = ''
```
colinvwood, line 515: I think this unintentionally overwrites `line` here -- dropping the large "FIXME" comment from above?

colinvwood: Actually looks like `_append_unknown_param` is used in other places than just above. So do we want to delete the large comment or move it into this method?

Oddant1: Good catch. This not being caught by tests means we do not have a test where an action that does exist in the env is run with a param that does not exist, probably wouldn't be too hard to bs one

colinvwood (approved):

colinvwood:
tested locally 👍🏻

PR #387, opened 2026-01-10
Adds warning filtering to stderr

Macabe222:
Adds a context manager `filter_rachis_warnings` that writes all `RachisWarnings` to stderr and allows others to pass to the redirected stdout/stderr.

### DIFF
--- a/q2cli/commands.py
+++ b/q2cli/commands.py
@@ -548,26 +548,26 @@ class ActionCommand(BaseCommandMixin, click.Command):
 
         cleanup_logfile = False
         try:
-            with filter_QIIME2_warnings():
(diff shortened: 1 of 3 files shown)


colinvwood:
needs qiime2/qiime2#898

q2cli/commands.py
```diff
@@ -14,6 +14,7 @@
 
 from q2cli.click.command import BaseCommandMixin
 from q2cli.core.config import CONFIG
+from q2cli.util import capture_rachis_warnings
```
Oddant1: This import can be moved down and put inside the method it's used in now I believe

PR #388, opened 2026-01-13
MAINT: Split up replay script test assertions

Oddant1:
Split up assertions of imports in templated replay script so order doesn't matter


lizgehret:
this passes for me locally and the change is effectively harmless so i'm going to go ahead and merge and ignore the CI failures.

PR #390, opened 2026-02-10
Usage examples now reference rachis not qiime2

Oddant1:
Usage examples now import from rachis not from qiime2 (at least I'm seeing failures due to this locally). This updates the checks in the cli to check for rachis not qiime2.

Is this what we want? I'm assuming we want to use the new name.


gregcaporaso:
@Oddant1 - do you know why the build failed?

ebolyen:
I'm a little confused how there is a diff here, because #389 happened and is in dev.

ebolyen:
Hey @Oddant1 you were just behind, I pushed a merge commit and the diff is now empty.

CI is still upset, but @lizgehret is looking into that.

PR #391, opened 2026-02-10
Cli side of validate usage output names

Oddant1:
Cli side of rachis-org/rachis#908.

~~Blocked by rachis-org/rachis#908~~

### DIFF
--- a/q2cli/core/usage.py
+++ b/q2cli/core/usage.py
@@ -98,6 +98,7 @@ class CLIUsageVariable(usage.UsageVariable):
         if self.var_type not in ('format', 'column'):
             cli_name = self.to_cli_name(cli_name)
 
+        cli_name = shlex.quote(cli_name)
         return cli_name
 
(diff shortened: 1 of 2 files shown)


q2cli/core/usage.py
```diff
@@ -98,6 +98,7 @@ def to_interface_name(self):
         if self.var_type not in ('format', 'column'):
             cli_name = self.to_cli_name(cli_name)
 
+        cli_name = shlex.quote(cli_name)
```
Oddant1, line 101: @ebolyen, I liked this idea, but do you think this is sufficient? It also almost feels redundant... better safe than sorry maybe?

ebolyen: I think this could reasonably be called defensive programming, it won't hurt anything and isn't slow.

ebolyen (approved):

ebolyen:
Is anything else needed for this?

Oddant1:
@ebolyen I don't believe so

PR #393, opened 2026-02-16
MAINT: fix failing test due to string quote mismatch

lizgehret:


conda-recipe/meta.yaml
```diff
@@ -33,8 +33,8 @@ test:
   imports:
   - q2cli
   commands:
-  - QIIMETEST= qiime --help
-  - QIIMETEST= py.test --pyargs q2cli
+  - Q2CLIDEV=1 QIIMETEST= qiime --help
+  - Q2CLIDEV=1 QIIMETEST= py.test --pyargs q2cli
```
ebolyen, line 37: This turned out to be a subtle problem with the CLI cache. I'll be fixing that properly soon, but this is fine for now.

PR #399, opened 2026-06-12
Adds tool: `redact-metadata`

Macabe222:
# Description

Adds a tool that removes all metadata files from a given artifact.
Linked to: rachis-org/rachis#941

# AI Disclosure

- [X] NO AI USED.
- [ ] AI USED.

### DIFF
--- a/q2cli/builtin/tools.py
+++ b/q2cli/builtin/tools.py
@@ -1772,7 +1772,7 @@ def redact_metadata(input_path, output_path):
     from q2cli.core.config import CONFIG
 
     artifact = Artifact.load(input_path)
-    artifact.redact_metadata()
+    artifact._archiver.redact_metadata()
     artifact.save(output_path)
 
     click.echo(
--- a/q2cli/builtin/tools.py
+++ b/q2cli/builtin/tools.py
@@ -1749,3 +1749,36 @@ def signature_verify(input_path, name):
     click.echo(CONFIG.cfg_style('success',
                                 f'Signature {name} on Result '
                                 f'{input_path} verified successfully.'))
+
+
+@tools.command(
+    name='redact-metadata',
+    short_help='Remove metadata from an artifact.',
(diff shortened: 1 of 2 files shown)


q2cli/builtin/tools.py
```diff
@@ -1749,3 +1749,36 @@ def signature_verify(input_path, name):
...
+    '--output-path',
+    required=True,
+    help='Path to save artifact with redacted metadata to.'
+)
+def redact_metadata(input_path, output_path):
+    from qiime2 import Artifact
+    from q2cli.core.config import CONFIG
+
+    artifact = Artifact.load(input_path)
+    artifact._archiver.redact_metadata()
```
Oddant1: Probably try/except this a la https://github.com/qiime2/q2cli/pull/399/changes#diff-9a0ee5a3ce54262998a961b4d9e91f0c21041b590c2bf5b784be2ca704b09be4L1743. 

Just makes the exception look nicer in the terminal, and this function has a known Exception when called on a `Result` without `Metadata` to redact.

Oddant1:
Can we also get a test for this? Probably test redacting some Metadata and have a test that raises the exception for calling the tool on an Artifact with no metadata.

q2cli/builtin/tools.py
```diff
@@ -1749,3 +1749,42 @@ def signature_verify(input_path, name):
...
+    help='Path to save artifact with redacted metadata to.'
+)
+def redact_metadata(input_path, output_path):
+    from qiime2 import Artifact
+    from q2cli.core.config import CONFIG
+
+    artifact = Artifact.load(input_path)
+
+    try:
+        artifact._archiver.redact_metadata()
```
ebolyen: Ah, we should still have a method on `artifact` for this, but it can call out to the archiver for the true implementation. This is similar to the other methods and properties which just forward the implementation along.

PR #400, opened 2026-06-12
IMP: `qiime info` qiime2 -> rachis name change & add parsl version

### LINKED ISSUE #927: update labels in `qiime info` report to reference rachis
In the `qiime info` report, we should update *System versions* to reference `rachis`. I imagine something like:

```
System versions
Python version: 3.12.13
rachis release: 2026.4
rachis version: 2026.4.0.dev0+27.gcb647ee
q2cli version: 2026.4.0.dev0+2.gcfcff79

Installed plugins
metadata: 2026.4.0.dev0
types: 2026.4.0.dev0+2.g3c86994

Application config directory
/opt/conda/envs/tiny-dev-2026-04-23/var/q2cli

Config
Config Source: /opt/conda/envs/tiny-dev-2026-04-23/etc/qiime2_config.toml

Getting help
To find help and learning resources, visit https://qiime2.org.
To get help with configuring and/or understanding QIIME 2 parallelization, visit https://use.qiime2.org/en/stable/references/parallel-configuration.html
```

lizgehret:
# Description

closes https://github.com/rachis-org/rachis/issues/927

# AI Disclosure

- [x] NO AI USED.
- [ ] AI USED.

### DIFF
--- a/q2cli/builtin/info.py
+++ b/q2cli/builtin/info.py
@@ -40,6 +40,7 @@ def _get_parsl_ver():
 
     return parsl_ver
 
+
 def _echo_plugins():
     import q2cli.core.cache
 
--- a/q2cli/builtin/info.py
+++ b/q2cli/builtin/info.py
@@ -28,15 +28,28 @@ def _echo_version():
 def _get_parsl_ver():
     import os
     import pathlib
+    import importlib.metadata
 
-    conda_env_prefix = os.environ.get('CONDA_PREFIX')
+    conda_env_prefix = os.environ.get('CONDA_PREFIX', '')
     conda_meta_path = pathlib.Path(conda_env_prefix) / 'conda-meta'
 
     parsl_ver = None
-    for file in conda_meta_path.iterdir():
-        if file.stem.startswith('parsl'):
-            # version is in the structure of: parsl-2026.2.23-pyhcf101f3_0
-            parsl_ver = file.stem.split('-', 2)[1]
+
+    # prefer the parsl version in conda env if available
+    if conda_meta_path.exists():
+        for file in conda_meta_path.iterdir():
+            if file.stem.startswith('parsl-'):
+                # version is in the structure of: parsl-2026.2.23-pyhcf101f3_0
+                parsl_ver = file.stem.split('-', 2)[1]
+                break
+
+    # fall back to any externally installed version if conda env not detectable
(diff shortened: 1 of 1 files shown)


q2cli/builtin/info.py
```diff
@@ -19,11 +19,28 @@ def _echo_version():
...
+    click.echo('rachis release: %s' % qiime2.__release__)
+    click.echo('rachis version: %s' % qiime2.__version__)
     click.echo('q2cli version: %s' % q2cli.__version__)
 
 
+def _get_parsl_ver():
+    import os
+    import pathlib
+
+    conda_env_prefix = os.environ.get('CONDA_PREFIX')
```
ebolyen:
```suggestion
    conda_env_prefix = os.environ.get('CONDA_PREFIX', '')
```

There's been a handful of users who don't use conda environments (or at least not visible ones), so we should make this tolerant of a missing `CONDA_PREFIX`.

q2cli/builtin/info.py
```diff
@@ -19,11 +19,28 @@ def _echo_version():
...
+    conda_meta_path = pathlib.Path(conda_env_prefix) / 'conda-meta'
+
+    parsl_ver = None
+    for file in conda_meta_path.iterdir():
```
ebolyen: If we add a guard to check if the path exists, then that should work with the above.

q2cli/builtin/info.py
```diff
@@ -19,11 +19,28 @@ def _echo_version():
     pyver = sys.version_info
     click.echo('Python version: %d.%d.%d' %
                (pyver.major, pyver.minor, pyver.micro))
-    click.echo('QIIME 2 release: %s' % qiime2.__release__)
-    click.echo('QIIME 2 version: %s' % qiime2.__version__)
+    click.echo('Parsl version: %s' % _get_parsl_ver())
```
ebolyen, line 22: I think we should try-except this, so that we report as a best-effort.

We could also probably use this stdlib method: https://docs.python.org/3/library/importlib.metadata.html#importlib.metadata.version

It won't be the conda version (I mean it probably will be), but I think that's about as good as our `qiime2.__version__` reporting.

ebolyen (changes requested):

ebolyen (approved):
Thanks @lizgehret!

PR #409, opened 2026-08-31
MAINT: qiime2/q2cli -> rachis/rachis-cli name change

lizgehret:
# Description

This updates all references of qiime/qiime2/q2cli to rachis/rachis-cli in the following locations:
 - doc/readme/help text references
 - base command call (i.e. `qiime` etc etc -> `rachis` etc etc)
 - script names & internal variable names (where applicable)
 - Rendered usage examples & provenance replay text

# AI Disclosure

- [ ] NO AI USED.
- [X] AI USED.

# AI Usage Details
Claude Opus 5 was used to assist with refactoring the tab completion script, and for doing a secondary repo-wide sweep to catch any name changes I missed.


rachis_cli/core/usage.py
```diff
@@ -130,8 +130,9 @@ def assert_output_type(self, semantic_type, key=None):
         if key:
             input_path = self._key_helper(input_path, key)
 
+        # TODO: update qiime-type to rachis-type ?
```
lizgehret:
```suggestion
```

rachis_cli/core/cache.py
```diff
@@ -145,8 +149,15 @@ def _get_current_requirements(self):
...
         # individual plugin names and versions in order to determine if the
         # cache is outdated.
-        for entry_point in importlib.metadata.entry_points(
-                group='qiime2.plugins'):
+        # Plugins are migrating from the `qiime2.plugins` entry point group
+        # to `rachis.plugins`. Read both, so that the cache still invalidates
+        # for plugins which have not migrated yet. This mirrors
+        # `rachis.sdk.PluginManager.iter_entry_points`. A distribution that
+        # declares both groups yields duplicate entry points, which the `set`
+        # above collapses.
```
ebolyen, line 159: Perfect!

rachis_cli/core/usage.py
```diff
@@ -88,12 +88,12 @@ def _key_helper(self, input_path, key):
         return "%s%s%s" % (input_path, key, self.ELEMENT_EXT[self.var_type])
 
     def to_interface_name(self):
-        if hasattr(self, '_q2cli_ref'):
-            return self._q2cli_ref
+        if hasattr(self, '_rachis_cli_ref'):
+            return self._rachis_cli_ref
```
ebolyen, line 92: Reminder to myself to update this in q2doc (and I think q2galaxy may care).

rachis_cli/__main__.py
```diff
@@ -8,37 +8,38 @@
...
+@click.command(cls=rachis_cli.commands.RootCommand,
+               invoke_without_command=True,
                no_args_is_help=True, help=ROOT_COMMAND_HELP)
-@click.version_option(prog_name='q2cli',
-                      message='%(prog)s version %(version)s\nRun `qiime info` '
-                              'for more version details.')
+@click.version_option(prog_name='rachis-cli',
+                      message='%(prog)s version %(version)s\nRun `rachis info`'
+                              ' for more version details.')
 def qiime():
```
ebolyen:
```suggestion
def rachis():
```

(but I think this will change your entrypoints in the pyproject toml slightly)

pyproject.toml
```diff
@@ -10,13 +10,18 @@ dynamic = ["version"]
 
 [project.urls]
 Homepage = "https://qiime2.org"
-Repository = "https://github.com/qiime2/q2cli"
+Repository = "https://github.com/rachis-org/rachis-cli"
 
 [project.scripts]
-"qiime" = "q2cli.__main__:qiime"
+"qiime" = "rachis_cli.__main__:qiime"
```
ebolyen: I'm not opposed to this, but I think it would be fine to make this a facet of `q2` compatibility package (just like `mosh`).

A consequence I guess would be you wouldn't `pip install` the `qiime` command in the future, but I feel ok with that I think.

@gregcaporaso do you have thoughts here?

gregcaporaso: I think not being able to `pip install qiime` is acceptable - is that what you mean?

ebolyen: Yep that's all, basically do we add the entry point here, or in a separate compatibility package. There may be another better reason to still add it to rachis_cli this that I am not seeing at the moment.

lizgehret: I don't have strong feelings about it; I think the goal of moving towards rachis in the language would align with not being able to pip install qiime (since this isn't a historical precedent, so people won't miss not being able to do this). And if we want the proxy packages to more closely mirror one another, it would make sense for this to live within q2 rather than rachis (since they're always shipped together in our distros anyways, this won't make any difference to our users).

lizgehret:
right, this is going to fail because i didn't use my own paired functionality correctly 😂

i'm going to re-open this on a separate paired branch so this can be tested with the rachis changes. i'll summarize remaining to-do's over there.

lizgehret:
superceded by #411

ebolyen (approved):
Code looks great to me! base-command handling seems like it's what we would want.

I am going to test this out on my linux laptop to double check Bash (and god knows what else), so I will do that shortly after I get home.

rachis_cli/tests/test_core.py
```diff
@@ -454,5 +455,21 @@ def test_replay_supplement_from_provdag(self):
                     self.assertIn(item, namelist_set)
 
 
+class BaseCommandTests(unittest.TestCase):
+    def setUp(self):
+        self.command = click.Command('root')
+
+    def test_installed_front_end(self):
+        front_ends = rachis_cli.util.get_cli_command_names()
+        self.assertTrue(front_ends)
```
ebolyen, line 464: 👍 that's a way easier way to prove the loop exists than anything I have ever done

ebolyen (approved):

